PR #73, opened 2021-05-05
fix plugin name registration

SwampDragons:
This plugin was registering the wrong names, which lead the Packer core never to discover and load the external plugin. This fixes the registration names.

Also add missing makefile targets and docs.zip elements from the goreleaser.

cc @nywilken


SwampDragons:
linter is failing because of all the calls to go fmt. I don't have the bandwidth to fix all these linting errors right now.

.goreleaser.yml
```diff
@@ -9,7 +9,7 @@ before:
     - go test ./...
     # As part of the release doc files are included as a separate deliverable for
     # consumption by Packer.io. To include a separate docs.zip uncomment the following command.
-    #- /bin/sh -c "[ -d docs ] && zip -r docs.zip docs/"
+    - /bin/sh -c "[ -d docs ] && zip -r docs.zip docs/"
```
sylviamoss, line 12: Fixing this to render partials
```suggestion
    - make ci-release-docs
```

PR #75, opened 2021-05-06
Fix plugin version

sylviamoss:
Moved version to main so that goreleaser can update those variables at release time.

### DIFF
--- a/version/version.go
+++ b/version/version.go
@@ -4,20 +4,4 @@ import (
 	"github.com/hashicorp/packer-plugin-sdk/version"
 )
 
-var (
-	// Version is the main version number that is being run at the moment.
-	Version = "0.0.1"
-
-	// VersionPrerelease is A pre-release marker for the Version. If this is ""
-	// (empty string) then it means that it is a final release. Otherwise, this
-	// is a pre-release such as "dev" (in development), "beta", "rc1", etc.
-	VersionPrerelease = "dev"
-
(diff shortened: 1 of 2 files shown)


version/version.go
```diff
@@ -4,20 +4,4 @@ import (
...
-
-	// PluginVersion is used by the plugin set to allow Packer to recognize
-	// what version this plugin is.
-	AzurePluginVersion *version.PluginVersion
-)
-
-func init() {
-	AzurePluginVersion = version.InitializePluginVersion(Version, VersionPrerelease)
-}
+var AzurePluginVersion *version.PluginVersion
```
sylviamoss, line 7: This is used internally and cannot be easily removed

nywilken:
The version information here is used by the client to create the UserAgent so I moved it out so that it can be managed by a separate version package. The .goreleaser file was also updated to account for the package. 

The current release panics with a nil pointer error. Is there a reason why it has to be in main?

sylviamoss:
Oh no, sorry for making those changes on top of yours. I looked at goreleaser and I saw this line https://github.com/hashicorp/packer-plugin-azure/blob/main/.goreleaser.yml#L28 which still uses the main package version and that's why I did it. 

We can revert this and update this specific goreleaser line to the correct package. But I kinda prefer the version in the main package as it's together with the plugin registration, which requires the version, and easier to find.

PR #79, opened 2021-05-07
add the "no-tty" flag to goreleaser sign

sarahethompson:
Adding as per - https://github.com/hashicorp/terraform-ls/pull/475

We can't re-produce this error when testing a similar configuration to this repo but worth a try!

```
gpg apparently uses tty for printing warnings (in this case likely about trust level for the new key)

Because there's no valid tty in GitHub Actions (or any other CI) we reflect that by passing extra flags to gpg.
```


azr (approved):
LGTM, thanks 🙂

PR #82, opened 2021-05-11
Reverting back to use github action for importing gpg key

sarahethompson:
See https://github.com/hashicorp/ghaction-import-gpg/pull/4 for the update to default-cache-ttl.


sylviamoss (approved):
Thanks!! 👍🏼

PR #89, opened 2021-05-18
Azure keyvault secret datasource

### LINKED ISSUE #83: Azure Keyvault Datasource
#### Community Note

#### Description

Add support for azure keyvault as a datasource, similarly to terraform azurerm keyvault datasource

#### Use Case(s)

* use any secret, key or certificate from keyvault in providers

#### Potential configuration

```
data "azurerm_key_vault_secret" "example" {
  name         = "secret-sauce"
  key_vault_id = "my-keyvault-id"
}
```

NicolasFloquet:
Hello,

So this is a first implementation of a datasource for keyvault secrets.
Configuration requires a keyvault name (I used "keyvault_id" to stay consistent with terraform's azure-arm), and a secret name. Datasource use the same authentication parameters than the builder:
Example:

```
data "azure-keyvault-secret" "secret" {
  name            = "my-secret"
  keyvault_id     = "my-keyvault"
  client_id       = "${var.client_id}"
  subscription_id = "${var.subscription_id}"
  client_secret   = "${var.client_secret}"
}
```

Output is also similar to what you get with terraform:
- `id` (string) - The Key Vault Secret ID.

- `value` (string) - The value of the Key Vault Secret.

- `content_type` (string) - The content type for the Key Vault Secret.

- `tags` (map[string]\*string) - Any tags assigned to this resource.

Closes #83


datasource/keyvaultsecret/data.go
```diff
@@ -0,0 +1,135 @@
...
+	}
+
+	return nil
+}
+
+func (d *Datasource) OutputSpec() hcldec.ObjectSpec {
+	return (&DatasourceOutput{}).FlatMapstructure().HCL2Spec()
+}
+
+// We need to stub packersdk ui "say" function. UI is not available from this package, I guess this is the best we can do for now
```
sylviamoss, line 83: this sounds like something we need to think about 🤔  nice workaround though!

sylviamoss (commented):
This is looking nice! I miss the docs though. I see you generated the partials but they are not being used anywhere.

NicolasFloquet:
Hello, 
This is my first contribution, so I have a few questions & comments. 

First of all, to avoid unnecessary code duplication for authentication, I reused some code from the builder (namely, `GetServicePrincipalTokens`). Yet, this methods relies on packer sdk UI "say" function, that is not available from the execute handler of the datasource. I stubbed the function with a log.Println, but I feel this is not ideal. Is there any way or plan to access UI functions from datasources?

You will also notice that I have not implemented any acceptance test for the datasource yet. I'm actually pretty confused about how it should work. I tried to look at the amazon plugin, and the acceptance seemingly requires credentials to pass. Can I get some guidance on how to implement this test?

Thank you!

NicolasFloquet:
> This is looking nice! I miss the docs though. I see you generated the partials but they are not being used anywhere.

Just added added a documentation page shamelessly inspired by the amazon one :)

sylviamoss:
Hey @NicolasFloquet, thanks for working on this. Again, this is looking nice! 

You're right, the UI is not available to the Datasource and it should be. This will require some thinking to find the best approach to do it. The first thing it comes to my mind is that we are going to change the Datasource interface to let `Execute()` get the UI as an argument. There are some ways to do it without breaking the existing interface. I will create an issue to track this and speak to the rest of the team about it. For now, I think your workaround should be good.

For the acceptance test, you can pass credentials via environment variables and document that these are needed in order to run the test. The amazon ones assume the credentials are set as env variables. Does that help you? 

Feel free to keep making any question you need 😃

NicolasFloquet:
Hello @sylviamoss,
I've added the documentation and the acceptance test.
For simplicity I chose to rely on env variables for the test. Following variables are required:
`PKR_VAR_keyvault_id`
`PKR_VAR_client_id`
`PKR_VAR_client_secret`
`PKR_VAR_tenant_id`

I think my pull request is ready for review now :)

docs/datasources/index.mdx
```diff
@@ -0,0 +1,30 @@
+---
+description: |
+  The Azure plugin is able to fetch data from Azure. To achieve this, the plugin comes with
+  data sources to retrieve keyvault secrets.
+page_title: Azure - Data Sources
+sidebar_title: Overview
```
sylviamoss, line 6: Sidebar_title works but is old, we are using nav_title now. 
```suggestion
nav_title: Overview
```

datasource/keyvaultsecret/data_acc_test.go
```diff
@@ -0,0 +1,124 @@
...
+	Id          string
+	ContentType string
+	Tags        map[string]*string
+
+	client keyvault.BaseClient
+}
+
+func getAuthorizer() (autorest.Authorizer, error) {
+	os.Setenv("AZURE_AD_RESOURCE", "https://vault.azure.net")
+	os.Setenv("AZURE_TENANT_ID", os.Getenv("PKR_VAR_tenant_id"))
```
nywilken, line 91: To simplify the acceptance testing setup the expected environment variables should be the same as expected by the current builder acceptance tests. This way you don't have to set anything, maybe other than the AZURE_AD_RESOURCE which looks like it can be a const for the test cases. 

https://github.com/hashicorp/packer-plugin-azure/blob/59806118c759250ad54715b6265e4b78b7dc5121/builder/azure/arm/builder_acc_test.go#L6

datasource/keyvaultsecret/test-fixtures/template.pkr.hcl
```diff
@@ -0,0 +1,47 @@
+
+variable "keyvault_id" {
+    type=string
+}
+variable "client_id" {
+    type=string
```
sylviamoss, line 6: I think these can default to the environment, is easier to have them set already than setting the PKR_VAR_client_id. What do you think? 
```suggestion
    type=string
    default = "${env("AZURE_CLIENT_ID")}"
```

datasource/keyvaultsecret/test-fixtures/template.pkr.hcl
```diff
@@ -0,0 +1,47 @@
...
+}
+variable "client_secret" {
+    type=string
```
sylviamoss, line 9:
```suggestion
    type=string
    default = "${env("AZURE_CLIENT_SECRET")}"
```

datasource/keyvaultsecret/test-fixtures/template.pkr.hcl
```diff
@@ -0,0 +1,47 @@
...
+}
+variable "tenant_id" {
+    type=string
```
sylviamoss, line 12:
```suggestion
    type=string
    default = "${env("AZURE_TENANT_ID")}"
```

datasource/keyvaultsecret/data_acc_test.go
```diff
@@ -0,0 +1,124 @@
...
+}
+
+type KeyvaultSecret struct {
+	Name        string
+	Value       string
+	Id          string
+	ContentType string
+	Tags        map[string]*string
+
+	client keyvault.BaseClient
```
nywilken, line 86: nit: Run go fmt against the files to format the source.

datasource/keyvaultsecret/test-fixtures/template.pkr.hcl
```diff
@@ -0,0 +1,47 @@
...
+}
+
+data "azure-keyvault-secret" "basic-example" {
+  name            = "packer-datasource-keyvault-test-secret"
+  keyvault_id     = "${var.keyvault_id}"
+  client_id       = "${var.client_id}"
+  client_secret   = "${var.client_secret}"
+  tenant_id       = "${var.tenant_id}"
```
sylviamoss, line 20: You can also just add a comment to this file telling that the `AZURE_*` envs are necessary to be set. If so you can remove the variables and these three lines, as well as their reference in `getAuthorizer()`

sylviamoss: I'm assuming azure consumes the env variables by default like AWS does. I think it does because I run my azure builds like this 🤔

datasource/keyvaultsecret/test-fixtures/template.pkr.hcl
nywilken: Is TENANT_ID still needed?

It has been removed from the other acceptance test as the ARM_SUBSCRIPTION_ID appears to be enough. But I'm not sure from the context of this change.

datasource/keyvaultsecret/data.go
nywilken: I agree with @sylviamoss that we need to think about the use of UI within datasources. But IMHO I don't think it is needed for this datasource.

 The use of the say function is solely for displaying the debug like messages from the GetServicePrincipalTokens function which can add some additional noise to the output since datasources execute before an actual build.

For this case I would suggestion going with the logSay function but with a few small tweaks. Which will only show up when running a Packer build with the debug logs enabled  `PACKER_LOG=1 packer build template.pkr.hcl`.

```go
func logSay(s string) {
  log.Println("[DEBUG] data.keyvaultsecret:", s)
}

**Review by nywilken — COMMENTED** [review]
@NicolasFloquet I agree with @sylviamoss this looks is looking good. I left a few additional suggestions and my thoughts on the logSay function workaround. I would like to pull this down and test as well. But at a glance it looks good.

File: datasource/keyvaultsecret/data.go (hunk 1 above)
Comment (NicolasFloquet) [reply datasource/keyvaultsecret/data.go:83]: Done

File: docs/datasources/index.mdx (hunk 2 above)
Comment (NicolasFloquet) [reply docs/datasources/index.mdx]: Done

File: datasource/keyvaultsecret/test-fixtures/template.pkr.hcl (hunk 6 above)
Comment (NicolasFloquet) [reply datasource/keyvaultsecret/test-fixtures/template.pkr.hcl:12]: Actually at first I wanted to stay with subscription Id, but the acceptance test uses azure-sdk-for-go for secret creation/deletion, and as far as I know the SDK only works with tenant Id (https://docs.microsoft.com/en-us/azure/developer/go/azure-sdk-authorization#use-environment-based-authentication). 
For simplicity purpose I also used tenant id in the template.

File: datasource/keyvaultsecret/test-fixtures/template.pkr.hcl (hunk 8 above)
Comment (NicolasFloquet) [reply datasource/keyvaultsecret/test-fixtures/template.pkr.hcl:20]: I tried Azure builder with only AZURE_* environment variable and it didn't work, it actually falls back to managed identity authentication, which obviously cannot work on my computer.
I'm genuinely curious to know how it works for you!
Having the same behavior than AWS plugin would be great indeed.

Comment (nywilken) [reply datasource/keyvaultsecret/test-fixtures/template.pkr.hcl:20]: The builder is setup to use managed identity if the user does not explicitly set subscription_id, client_id, and client_secret. Which is not configured automatically when using the AZURE_* environment variables. We run our builds and tests with the following variables:

```hcl
    "variables": {
        "client_id": "{{ env `ARM_CLIENT_ID` }}",
        "client_secret": "{{ env `ARM_CLIENT_SECRET` }}",
        "object_id": "{{ env `ARM_OBJECT_ID` }}",
        "subscription_id": "{{ env `ARM_SUBSCRIPTION_ID` }}",
        "tenant_id": "{{ env `ARM_TENANT_ID` }}",
        "resource_group_name": "{{ env `ARM_RESOURCE_GROUP_NAME` }}",
        "storage_account": "{{ env `ARM_STORAGE_ACCOUNT` }}"
    },
```

datasource/keyvaultsecret/data_acc_test.go
nywilken: After your last comment about the AZURE_* variables I understand why you have this setup. Instead of PKR_VAR_* I recommend using the environment variables already required for the the acceptance tests. For example

```
os.Setenv("AZURE_TENANT_ID", os.Getenv("ARM_TENANT_ID"))
os.Setenv("AZURE_CLIENT_ID", os.Getenv("ARM_CLIENT_ID"))
...
```

datasource/keyvaultsecret/test-fixtures/template.pkr.hcl
nywilken: Ah okay I see. That is why `os.Setenv("AZURE_AD_RESOURCE", "https://vault.azure.net")` is also needed. Thanks for the response here TIL.

datasource/keyvaultsecret/test-fixtures/template.pkr.hcl
NicolasFloquet: Okay, this is way clearer! I'll push a change match this setup then :)

nywilken: 👍 For the record this is an example in JSON. An HCL2 equivalent would be something like

```
variable "client_id" {
  type    = string
  default = "${env("ARM_CLIENT_ID")}"
}

variable "client_secret" {
  type    = string
  default = "${env("ARM_CLIENT_SECRET")}"
}

variable "object_id" {
  type    = string
  default = "${env("ARM_OBJECT_ID")}"
}

variable "resource_group_name" {
  type    = string
  default = "${env("ARM_RESOURCE_GROUP_NAME")}"
}

variable "storage_account" {
  type    = string
  default = "${env("ARM_STORAGE_ACCOUNT")}"
}

variable "subscription_id" {
  type    = string
  default = "${env("ARM_SUBSCRIPTION_ID")}"
}

variable "tenant_id" {
  type    = string
  default = "${env("ARM_TENANT_ID")}"
}```

PR #115, opened 2021-07-03
Update url for example dir in arm.mdx

lyzlisa:
The link for examples in the ARM documentation was returning HTTP 404 code.

This PR updates the so that it no longer returns 404.


nywilken (approved):
Thanks for the fix.

PR #138, opened 2021-08-27
Add HCP Packer registry image metadata to builder artifacts

nywilken:


builder/azure/arm/artifact.go
```diff
@@ -249,13 +250,45 @@ func (*Artifact) Destroy() error {
...
+	if a.isManagedImage() {
+		id = a.ManagedImageId
+		location = a.ManagedImageLocation
+
+		labels := make(map[string]interface{})
+		labels["os_type"] = a.OSType
+		labels["managed_image_resourcegroup_name"] = a.ManagedImageResourceGroupName
+		labels["managed_image_name"] = a.ManagedImageName
+		if a.ManagedImageSharedImageGalleryId != "" {
+			labels["managed_image_sharedimagegallery_id"] = a.ManagedImageSharedImageGalleryId
```
SwampDragons, line 265: We probably also need the SIG name, as well as the SIG Image Name, and the SIG version.

SwampDragons (approved):
Super solid. We need a few more labels for images that are associated with a SIG, noted inline. Otherwise, this works as expected! (Check the vendoring; I had to change the SDK versioning to v0.2.5 to get it to build.

azr (approved):
LGTM

builder/azure/arm/artifact.go
nywilken: See PR #150

PR #161, opened 2021-11-02
Converged (a bit) to azure-arm coding style, added sysprep_done flag …

flixman:
Sometimes, on win images, sysprep does not work as expected. The recommendation then is to perform the sysprep manually (removing the agent extensions, then calling sysprep, etc.), in which case packer can skip the sysprep. This PR provides that functionality and also performs some small refactoring to make the azure-arm and azure-dtl builders more alike.

### DIFF
--- a/builder/azure/dtl/config.go
+++ b/builder/azure/dtl/config.go
@@ -525,7 +525,7 @@ func setUserNamePassword(c *Config) error {
 		c.Password = c.tmpAdminPassword
 	}
 
-        return nil
+	return nil
 }
 
 func provideDefaultValues(c *Config) error {
--- a/builder/azure/dtl/config.go
+++ b/builder/azure/dtl/config.go
@@ -20,6 +20,7 @@ import (
 	"github.com/Azure/go-autorest/autorest/to"
 	"github.com/masterzen/winrm"
 
+	azcommon "github.com/hashicorp/packer-plugin-azure/builder/azure/common"
 	"github.com/hashicorp/packer-plugin-azure/builder/azure/common/client"
 	"github.com/hashicorp/packer-plugin-azure/builder/azure/common/constants"
 
@@ -278,6 +279,9 @@ type Config struct {
 	// DisallowPublicIPAddress - Indicates whether the virtual machine is to be created without a public IP address.
 	DisallowPublicIP bool `mapstructure:"disallow_public_ip" required:"false"`
 
+        // SysPrepDone - Indicates whether SysPrep is to be requested to the DTL or has been already applied.
+        SysPrepDone bool `mapstructure:"sysprep_done" required:"false"`
+
 	// Runtime Values
 	UserName                string `mapstructure-to-hcl2:",skip"`
 	Password                string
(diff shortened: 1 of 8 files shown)


flixman:
@sylviamoss Excuse me for referencing you directly but... what do you think about this PR? :-)

flixman:
@nywilken Sorry for referencing you directly, but ... seems my PR is completely ignored? :-/

nywilken:
Hi @flixman no need for apologies and thanks for the ping. We've been a bit heads down on some Packer core related changes so we are working through the pending PRs for each of the plugins with a bit of delay. So our apologies there. 

I wanted to review your PR on Friday but just got caught up with a bug for Azure. I will be taking a look as I would like to try and pull  in your refactor before making the fix I have planned. Thanks for the contribution.

builder/azure/dtl/config.go
```diff
@@ -278,6 +279,9 @@ type Config struct {
 	// DisallowPublicIPAddress - Indicates whether the virtual machine is to be created without a public IP address.
 	DisallowPublicIP bool `mapstructure:"disallow_public_ip" required:"false"`
 
+	// SysPrepDone - Indicates whether SysPrep is to be requested to the DTL or has been already applied.
+	SysPrepDone bool `mapstructure:"sysprep_done" required:"false"`
```
nywilken, line 283: I found the name for this configuration to be confusing. From the description I take it to mean that if set to true the source image already had sysprep applied to it so there is no need to rerun again. Is that correct? 
```go
	// SkipSysprep - Indicates whether SysPrep is to be requested to the DTL or if it should be skipped because it has already been applied. Defaults to false.
	SkipSysprep bool `mapstructure:"skip_sysprep" required:"false"`
```

If my understanding is correct, I would suggest calling this argument "skip_sysprep"

nywilken (changes requested):
Overall this looks good. Thanks for the refactor and contribution. 

I left one suggestion but this should be otherwise good to go. I'm running the acceptance tests. I will report back if there are any failures.

builder/azure/dtl/config.go
flixman: good point! done!

nywilken:
Hi @flixman thanks for making the update. It looks like you need to regenerate the HCL configuration after you latest change. You can fix this by running `make generate` or `go generate ./...` if you don't have Make installed, then commit the generated files. 


I tried testing this code and the code on main but I'm running into an issue, which I pasted below. I don't have too much experience with DTL so it might just be my setup. Can you confirm that you are able to build as expected using the updated code? 

```
2022/02/07 18:28:33 ui error: ==> azure-dtl: dtl.VirtualMachinesClient#ApplyArtifacts: Failure sending request: StatusCode=0 -- Original Error: Code="ApplyArtifactsUnavailable" Message="Apply artifacts feature is currently unavailable. This can occur when the virtual machine is stopped or the agent is unresponsive. If the issue persists, refer to the troubleshooting guide (https://aka.ms/ArtifactsTroubleshoot)."
```

nywilken (approved):
@flixman thanks again for the contribution. This is good to go. 

 I committed two small changes to address the documentation generation and fix the failing tests. The test failure was due to a timeout issue; not with the changes in the code.

flixman:
@nywilken Sorry for the radio silence: I have been catching fire for several months, and this felt off my mental stack :-/. Thank you!!!

PR #162, opened 2021-11-02
Fix mistyped builder in website's documentation

JKalliomaki:
There is a mistyped type for builder type in the documentation. It says the builder type is "amazon-arm" instead of "azure-arm"

This is my first pull request and I'm making it via GitHub's _Propose changes_ tool so I hope this pull request will be up to standards.


sylviamoss (approved):
Nice catch! Thanks for fixing it 😄  👍🏼

PR #184, opened 2022-02-11
Add support for running the chroot builder from a VM scale set

szamfirov:
### Description

With the current implementation it is impossible to build VM images from a VM scale set machine using the chroot builder. The following PR adds support for VM scale set virtual machines by making use of the `VirtualMachineScaleSetVMsClient`. The original request for this functionality can be found here: https://github.com/hashicorp/packer/issues/9348, which is identical to our own use case.

### Changes in detail

The following changes have been made:
* Adding the `VirtualMachineScaleSetVMsClient`
* Adding helper functions to: get VM details, get VM disks, set VM disks using the new client
* The existing logic will fall back to the new implementation in case of errors - it will first try to get details as if running from a standalone VM and if any of these calls fail, it will use the `VirtualMachineScaleSetVMsClient` to retry and continue with the rest of the logic.
* Fix unmounting of filesystem on success - multiple filesystem mountpoints are initially mounted under the same location which was causing issues with the final cleanup step. The unmount is now done recursively which solves this issue (see [-R, --recursive](https://man7.org/linux/man-pages/man8/umount.8.html) from the official manual pages). Recursion for each directory will stop if any unmount operation in the chain fails for any reason (and thus preserving the same behavior).

### Risks associated with the changes

**No risks** - there is no negative effect on the existing implementation. We are already using the changes in anger and we haven't discovered any issues.

### Related open issue(s):

Closes [#9348](https://github.com/hashicorp/packer/issues/9348)

### DIFF
--- a/builder/azure/chroot/diskattacher.go
+++ b/builder/azure/chroot/diskattacher.go
@@ -15,9 +15,9 @@ import (
 )
 
 type DiskAttacher interface {
-	AttachDisk(ctx context.Context, disk string) (lun int32, err error)
+	AttachDisk(ctx context.Context, state multistep.StateBag, disk string) (lun int32, err error)
 	WaitForDevice(ctx context.Context, i int32) (device string, err error)
-	DetachDisk(ctx context.Context, disk string) (err error)
+	DetachDisk(ctx context.Context, state multistep.StateBag, disk string) (err error)
 	WaitForDetach(ctx context.Context, diskID string) error
 }
 
--- a/builder/azure/chroot/diskattacher.go
+++ b/builder/azure/chroot/diskattacher.go
@@ -10,20 +10,20 @@ import (
 	"github.com/Azure/azure-sdk-for-go/services/compute/mgmt/2019-12-01/compute"
 	"github.com/Azure/go-autorest/autorest/to"
 	"github.com/hashicorp/packer-plugin-azure/builder/azure/common/client"
-	"github.com/hashicorp/packer-plugin-sdk/multistep"
 	packersdk "github.com/hashicorp/packer-plugin-sdk/packer"
 )
 
 type DiskAttacher interface {
-	AttachDisk(ctx context.Context, state multistep.StateBag, disk string) (lun int32, err error)
+	AttachDisk(ctx context.Context, disk string) (lun int32, err error)
 	WaitForDevice(ctx context.Context, i int32) (device string, err error)
-	DetachDisk(ctx context.Context, state multistep.StateBag, disk string) (err error)
+	DetachDisk(ctx context.Context, disk string) (err error)
 	WaitForDetach(ctx context.Context, diskID string) error
 }
 
-var NewDiskAttacher = func(azureClient client.AzureClientSet) DiskAttacher {
+var NewDiskAttacher = func(azureClient client.AzureClientSet, ui packersdk.Ui) DiskAttacher {
 	return &diskAttacher{
 		azcli: azureClient,
+		ui:    ui,
 	}
 }
 
@@ -31,19 +31,18 @@ type diskAttacher struct {
 	azcli client.AzureClientSet
 
 	vm *client.ComputeInfo // store info about this VM so that we don't have to ask metadata service on every call
+	ui packersdk.Ui
 }
 
 var DiskNotFoundError = errors.New("Disk not found")
 
-func (da *diskAttacher) DetachDisk(ctx context.Context, state multistep.StateBag, diskID string) error {
-	ui := state.Get("ui").(packersdk.Ui)
-
+func (da *diskAttacher) DetachDisk(ctx context.Context, diskID string) error {
 	log.Println("Fetching list of disks currently attached to VM")
 	currentDisks, err := da.getDisks(ctx)
 	if err != nil {
 		log.Printf("DetachDisk.getDisks: error: %+v\n", err)
 		log.Println("Checking to see if instance is part of a VM scale set before giving up.")
-		ui.Say("Initial call for fetching VM instance disks returned an error. Checking to see if instance is part of a VM scale set before giving up.")
+		da.ui.Say("Initial call for fetching VM instance disks returned an error. Checking to see if instance is part of a VM scale set before giving up.")
 		currentDisks, err = da.getScaleSetDisks(ctx)
 		if err != nil {
 			return err
@@ -67,7 +66,7 @@ func (da *diskAttacher) DetachDisk(ctx context.Context, state multistep.StateBag
 	if err != nil {
 		log.Printf("DetachDisk.setDisks: error: %+v\n", err)
 		log.Println("Checking to see if instance is part of a VM scale set before giving up.")
-		ui.Say("Initial call for setting VM instance disks returned an error. Checking to see if instance is part of a VM scale set before giving up.")
+		da.ui.Say("Initial call for setting VM instance disks returned an error. Checking to see if instance is part of a VM scale set before giving up.")
 		err = da.setScaleSetDisks(ctx, newDisks)
 		if err != nil {
 			return err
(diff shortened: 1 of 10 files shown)


builder/azure/chroot/diskattacher.go
```diff
@@ -38,7 +37,10 @@ func (da *diskAttacher) DetachDisk(ctx context.Context, diskID string) error {
 	log.Println("Fetching list of disks currently attached to VM")
 	currentDisks, err := da.getDisks(ctx)
 	if err != nil {
-		return err
+		currentDisks, err = da.getScaleSetDisks(ctx)
+		if err != nil {
+			return err
+		}
```
nywilken, line 43: Is there a way to identify the exact error that can be ignored from getDisk? 

With this implementation we will lose the original error from calling da.getDisks(). It would be great to have both errors in the case that both function calls fail.

nywilken (commented):
Overall this looks good. All the acceptance test pass but I left a comment around losing potential error information. Thanks for making the fix. Please let us know if you have any questions on the review.

builder/azure/chroot/diskattacher.go
szamfirov: Apologies for the lack of activity but I was quite stretched this week.

I agree that we should bubble the error up somehow for the user to be aware but I'm struggling to find a proper solution. With the current implementation, any error (or MultiError for that matter) which is returned from DetachDisk, AttachDisk and the likes will result in a `multistep.ActionHalt` in the main "step" which will essentially stop the build process. So this might not be the perfect solution.

Is there a decent way to surface such message as a "warning"? I'm thinking of something much simpler which is just to notify the user that the initial call failed and display what the error was before the falling back to the VMSS logic.

Let me share an example for a bit better understanding. Something along the lines of:
```
currentDisks, err := da.getDisks(ctx)
if err != nil {
  log.Printf("DetachDisk.getDisks: error: %+v\nFalling back to the VM scale set implementation...", err)
  // We can even make this explicit in the UI
  ui.Say("Fetching VM instance disks failed. Enable debugging for more details. Falling back to the VM scale set implementation...")
  currentDisks, err = da.getScaleSetDisks(ctx)
  if err != nil {
    return err
  }
}

Comment (szamfirov) [reply builder/azure/chroot/diskattacher.go:43]: @nywilken, please let me know what you think on the above ☝️

**Comment by szamfirov** [comment]
@nywilken, just a friendly nudge. Please let me know what you think on the above ☝️

**Comment by szamfirov** [comment]
Hi @nywilken, thank you for the detailed response.

> Or are you suggesting that we don't return immediately after the first getDisk error, warn the user that the first call failed, and retry the getDisk call but for VMSS?
That's exactly what I'm suggesting. Otherwise it would be difficult to determine when will be the right time to proceed with the fallback to VMSS.

I've updated the PR and have added more detailed logging (both UI and debug logs). Please have a look when you have the time.

**Comment by nywilken** [comment]
@szamfirov apologies for the delay. I've been away from the machines for the past few weeks (vacation and things). I'll get to this PR next week. Thanks for updating the PR and for keeping a pulse on the changes. Excited to help get these changes in.

**Comment by Chivolta** [comment]
Hi guys. Any news?

File: builder/azure/chroot/diskattacher.go [hunk 2]
```diff
@@ -8,15 +8,16 @@ import (
...
-	"github.com/Azure/go-autorest/autorest/azure"
 	"github.com/Azure/go-autorest/autorest/to"
 	"github.com/hashicorp/packer-plugin-azure/builder/azure/common/client"
+	"github.com/hashicorp/packer-plugin-sdk/multistep"
+	packersdk "github.com/hashicorp/packer-plugin-sdk/packer"
 )
 
 type DiskAttacher interface {
-	AttachDisk(ctx context.Context, disk string) (lun int32, err error)
+	AttachDisk(ctx context.Context, state multistep.StateBag, disk string) (lun int32, err error)
```
nywilken, line 18: As opposed to passing in the entire statebag to get to the UI. I would recommend adding a field to each of these types for setting the UI. This way you can access the UI simply by calling `da.ui.(....)`. 

For example given the DiskAttacher type you could add the field
```go
type diskAttacher struct {
	azcli client.AzureClientSet
	vm *client.ComputeInfo // store info about this VM so that we don't have to ask metadata service on every call
        ui packersdk.Ui
}

var NewDiskAttacher = func(azureClient client.AzureClientSet, ui packersdk.Ui) DiskAttacher {
	return &diskAttacher{
		azcli: azureClient,
                ui: ui
	}
}

func (da *diskAttacher) DetachDisk(ctx context.Context, diskID string) error {
 //...

	log.Println("Fetching list of disks currently attached to VM")
	currentDisks, err := da.getDisks(ctx)
	if err != nil {
....
		da.ui.Say("Initial call for fetching VM instance disks returned an error. Checking to see if instance is part of a VM scale set before giving up.")

```
Then in the using the unit test as an example you could write
```go

	testDiskName := t.Name()

	errorBuffer := &strings.Builder{}
	ui := &packersdk.BasicUi{
		Reader:      strings.NewReader(""),
		Writer:      ioutil.Discard,
		ErrorWriter: errorBuffer,
	}

	state := new(multistep.BasicStateBag)
	state.Put("azureclient", azcli)
        da := NewDiskAttacher(azcli, ui)
```

TBH it feels a little weird to me to have all these exported fields and would prefer to have Ui and AZcli as the field names. But I think sticking with the style that exists is fine for now.

nywilken (changes requested):
I find the updated approach to work. I will pull down the changes and give the PR a run. In the meantime, I left a few suggestions for how to pass in a Ui for each of the Attacher types.

builder/azure/chroot/diskattacher.go
szamfirov: Fully agree with you - passing the state felt a bit too much. Thanks for the suggestion. I managed to incorporate the new field in my latest commit ([395277a](https://github.com/hashicorp/packer-plugin-azure/pull/184/commits/395277ae7283f2930fa68eb4a94fa06ee8da32c9)). It should all be fine now.

szamfirov:
@nywilken, please let me know once you manage to give these changes a spin. I'm looking forward to having this merged.

builder/azure/chroot/diskattacher.go
```diff
@@ -56,7 +64,13 @@ func (da *diskAttacher) DetachDisk(ctx context.Context, diskID string) error {
...
 	err = da.setDisks(ctx, newDisks)
 	if err != nil {
-		return err
+		log.Printf("DetachDisk.setDisks: error: %+v\n", err)
+		log.Println("Checking to see if instance is part of a VM scale set before giving up.")
+		da.ui.Say("Initial call for setting VM instance disks returned an error. Checking to see if instance is part of a VM scale set before giving up.")
+		err = da.setScaleSetDisks(ctx, newDisks)
+		if err != nil {
+			return err
+		}
```
nywilken, line 73: So I've walked down this path a bit because it seems like we could be continuing when we should error. Thinking in the case where the VM is not in a scale set and da.setDisks returns an error but for some strange reason the call to `da.setScaleSetDisks()` doesnt. The the original error which was printed to the screen would not return and allow the build to continue up until something downstream fails because the attach disk is not present. 

This case seems unlikely to me as I would expect `da.setScaleSetDisks` to always error when the vm is not in a scale set. But I did want to mention it because it was not immediately clear that something downstream checks for the existence of the said disks and fail appropriately.

builder/azure/chroot/diskattacher_test.go
```diff
@@ -17,9 +21,21 @@ import (
...
+	ui := &packersdk.BasicUi{
+		Reader:      strings.NewReader(""),
+		Writer:      ioutil.Discard,
+		ErrorWriter: errorBuffer,
+	}
+
+	state := new(multistep.BasicStateBag)
+	state.Put("azureclient", azcli)
+	state.Put("ui", ui)
+
```
nywilken, line 36: State is no longer needed
```suggestion
```

nywilken (approved):
Thanks for your patience in getting this change in. The changes look good to me. I left a comment around my thinking when walking through the multiple error logic but I don't think it is an issue.  You might have more insight as you are actually using this change already. Please let me know if you have thoughts.

builder/azure/chroot/diskattacher_test.go
```diff
@@ -2,12 +2,16 @@ package chroot
...
 	"context"
+	"io/ioutil"
+	"strings"
 	"testing"
 
 	"github.com/Azure/go-autorest/autorest/to"
 
 	"github.com/Azure/azure-sdk-for-go/services/compute/mgmt/2019-12-01/compute"
 	"github.com/hashicorp/packer-plugin-azure/builder/azure/common/client"
+	"github.com/hashicorp/packer-plugin-sdk/multistep"
```
nywilken, line 13: No longer needed
```suggestion
```

szamfirov:
@nywilken, thank you for merging the changes. I'm curious when the next release will be carved out so we can switch to using the official plugin (as opposed to the forked version)? Thanks in advance!

nywilken:
> @nywilken, thank you for merging the changes. I'm curious when the next release will be carved out so we can switch to using the official plugin (as opposed to the forked version)? Thanks in advance!

I'm working on getting out a release this week. If not today (EST) then tomorrow. There is an update to the Packer plugin sdk that I would like to get into the next release.

PR #186, opened 2022-02-20
feat: support skipping image creation

jkoelker:
Adds the config option `skip_create_image` to the builders to support
skipping image creation.

Fixes: #180

### DIFF
--- a/builder/azure/common/config.go
+++ b/builder/azure/common/config.go
@@ -1,0 +1,33 @@
+//go:generate packer-sdc struct-markdown
+
+package common
+
+import (
+	"github.com/hashicorp/packer-plugin-sdk/multistep"
+)
+
+const (
+	SkippingImageCreation = "Skipping image creation..."
+)
+
+type Config struct {
+	// Skip creating the image.
+	// Useful for setting to `true` during a build test stage.
+	// Defaults to `false`.
+	SkipCreateImage bool `mapstructure:"skip_create_image" required:"false"`
+}
+
+// CaptureSteps returns the steps unless `SkipCreateImage` is `true`. In that case it returns
+// a step that to inform the user that image capture is being skipped.
+func (config Config) CaptureSteps(say func(string), steps ...multistep.Step) []multistep.Step {
+	if !config.SkipCreateImage {
+		return steps
+	}
+
+	return []multistep.Step{
+		&StepNotify{
+			message: SkippingImageCreation,
+			say:     say,
+		},
+	}
+}
--- a/builder/azure/common/config_test.go
+++ b/builder/azure/common/config_test.go
@@ -1,0 +1,44 @@
+package common_test
+
+import (
+	"context"
+	"testing"
+
+	"github.com/hashicorp/packer-plugin-sdk/multistep"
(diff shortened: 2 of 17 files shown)


builder/azure/common/config.go
```diff
@@ -0,0 +1,33 @@
...
+)
+
+const (
+	SkippingImageCreation = "Skipping image creation..."
+)
+
+type Config struct {
+	// Skip creating the image.
+	// Useful for setting to `true` during a build test stage.
+	// Defaults to `false`.
```
msdolbey, line 16: Should the default be first?

```suggestion
	// Defaults to `false`.
	// Useful for setting to `true` during a build test stage.
```

msdolbey (approved):
👍

builder/azure/common/config.go
jkoelker: It looks like in the rest of the documentation the defaults is the last thing mentioned. I think we should keep this new one consistent, but would be happy to reorg it if desired by the maintainers.

msdolbey: Oh, I just saw it different in some other spots - but def trust you on this one - we can resolve this.

jkoelker:
Please let me know if this shared step approach is desirable. For CI purposes, I really only need something for the `azure-arm` builder, and if desired I can narrow the scope to only skipping the image capture for that builder.

jkoelker:
I looked into the CI failures and it looks like they are existing in the repo due to commented out code. Please let me know if you want me to fix them. I took a stab at it briefly, but realized that the commented code and related issues with it may be part of a larger plan that I'm unaware of.

nywilken:
Thanks for opening up this PR. We'll work on getting this reviewed as soon as possible to get you feedback on the changes.

garimakhulbe02:
@jkoelker any update here? We also need this merge :)

garimakhulbe02:
@nywilken any update? We need this merge too :)

garimakhulbe02:
@nywilken Can you prioritize this PR? We have customers block on this.

builder/azure/common/config_test.go
```diff
@@ -0,0 +1,44 @@
...
+	"context"
+	"testing"
+
+	"github.com/hashicorp/packer-plugin-sdk/multistep"
+	"github.com/stretchr/testify/assert"
+
+	"github.com/hashicorp/packer-plugin-azure/builder/azure/common"
+)
+
+func TestSkipCreateImage(t *testing.T) {
```
JenGoldstrich, line 13: I think we should add a test case here where we pass in false for SkipCreateImage, but this is good for now.

JenGoldstrich (approved):
Hi @jkoelker and all,

I've spent a bit of time today getting up to speed on this PR, testing it for myself, and thinking about the change.

Overall it's a very good, kudos @jkoelker! I pointed out a place where we could add another unit test, and I think it would be good to include some example packer template files for this field, but I do not want to block merging of this PR on either of those issues.

I will look into back filling those two things soon, but would of course appreciate a PR adding them.

Merging.

garimakhulbe02:
@jkoelker @JenGoldstrich Thank you!

PR #200, opened 2022-05-20
Address linter issues

JenGoldstrich:
Golang CI has been failing for several months, a lot of the issues I fixed were more than 6-7 years old so I don't believe anything specific changed in the code base to make lint fail, but rather golang-ci lint added new rules at some point.

I tried to keep this relatively simple, any linting changes that seemed like they might have larger code complications I avoided and added a TODO (and in one case a nolint)

### DIFF
--- a/builder/azure/arm/builder_acc_test.go
+++ b/builder/azure/arm/builder_acc_test.go
@@ -20,21 +20,14 @@ package arm
 //   go test -v -timeout 90m -run TestBuilderAcc_.*
 
 import (
-	"bytes"
-	"context"
 	_ "embed"
-	"errors"
 	"fmt"
 	"io/ioutil"
 	"os"
 	"os/exec"
-	"strings"
 	"testing"
 
-	"github.com/Azure/go-autorest/autorest"
-	"github.com/Azure/go-autorest/autorest/azure/auth"
 	"github.com/hashicorp/packer-plugin-sdk/acctest"
-	packersdk "github.com/hashicorp/packer-plugin-sdk/packer"
 )
 
 const DeviceLoginAcceptanceTest = "DEVICELOGIN_TEST"
@@ -190,7 +183,10 @@ func TestBuilderAcc_Blob_Windows(t *testing.T) {
 
 func TestBuilderAcc_Blob_Linux(t *testing.T) {
 	b := Builder{}
-	b.Prepare()
+	_, _, err := b.Prepare()
+	if err != nil {
+		t.Fatal(err)
+	}
 	acctest.TestPlugin(t, &acctest.PluginTestCase{
 		Name:     "test-azure-blob-linux",
 		Type:     "azure-arm",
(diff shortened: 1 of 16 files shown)


builder/azure/arm/template_factory.go
```diff
@@ -92,7 +92,10 @@ func GetVirtualMachineDeployment(config *Config) (*resources.Deployment, error)
 			config.ImageSku,
 			config.ImageVersion)
 
-		builder.SetManagedMarketplaceImage(config.Location, config.ImagePublisher, config.ImageOffer, config.ImageSku, config.ImageVersion, imageID, config.managedImageStorageAccountType, config.diskCachingType)
+		err = builder.SetManagedMarketplaceImage(config.Location, config.ImagePublisher, config.ImageOffer, config.ImageSku, config.ImageVersion, imageID, config.managedImageStorageAccountType, config.diskCachingType)
```
sylviamoss, line 95: could this be a breaking change?  I don't know the context of `SetManagedMarketplaceImage` but what happened before when this failed? 🤔

sylviamoss (approved):
I think this looks good. 👍🏼  
I raised two concerns and I'd like to know what you think.

builder/azure/arm/builder_acc_test.go
```diff
@@ -265,96 +257,6 @@ func TestBuilderAcc_rsaSHA2OnlyServer(t *testing.T) {
 	})
 }
 
-func testAccPreCheck(*testing.T) {}
```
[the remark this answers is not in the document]
JenGoldstrich: That's good context, I can re-add these methods and just add //nolint to them so that hte linter doesn't fail for these unused functions

builder/azure/arm/template_factory.go
JenGoldstrich: https://github.com/hashicorp/packer-plugin-azure/blob/6c11375deca7709accbc00403181db532be99cd7/builder/azure/common/template/template_builder.go#L157

It looks like this method only fails when a specific field isn't set in the template so it might make sense for it to fail here, but I think it's worth changing this back just so we aren't making a potential breaking change (even a logically correct one) in this PR

builder/azure/arm/step_get_ip_address.go
```diff
@@ -73,7 +73,8 @@ func (s *StepGetIPAddress) getPublicIP(ctx context.Context, resourceGroupName st
 }
 
 func (s *StepGetIPAddress) getPublicIPInPrivateNetwork(ctx context.Context, resourceGroupName string, ipAddressName string, interfaceName string) (string, error) {
-	s.getPrivateIP(ctx, resourceGroupName, ipAddressName, interfaceName)
+	// TODO: This was being called without capturing any return variables, causing linter issue, as far as I can tell calling getPrivateIP here does nothing, look into removing
+	_, _ = s.getPrivateIP(ctx, resourceGroupName, ipAddressName, interfaceName)
```
nywilken, line 77: Should we just remove this call? It looks like this method does is not making any changes to the step via some pointer modification and that the return values are not used. 
```suggestion

```

builder/azure/chroot/diskattacher.go
```diff
@@ -123,6 +123,8 @@ func (da *diskAttacher) AttachDisk(ctx context.Context, diskID string) (int32, e
 
 	// disk was not found on VM, go and actually attach it
 
+	// TODO This assignment looks like it would do nothing, consider removing it
```
nywilken, line 126: The variable is being used after the for loop, which is while it is being defined before the label. The logic is a little confusing. I think this could benefit from a refactor. 
```suggestion
	// TODO This logic is a little hard to read; refactor to remove the label. 
```

nywilken (approved):
Changes looks goo. I left two suggestions

PR #202, opened 2022-05-26
Update Availability Zones list and fix warning message

maxilampert:
This PR fixes the following:

- Update the list of supported Availability zones based on https://docs.microsoft.com/en-us/azure/availability-zones/

- Fix the warning message to work properly in cases where a user did specify the location in the exact notation in the template


nywilken (approved):
Changes look great. Thanks for the update.

PR #220, opened 2022-08-03
Support Trusted Launch VMs

maxilampert:
Sets the SecurityProfile if either Secure Boot or the VTPM option has been set in the packer configuration. This allows the deployment of VMs using Trusted Launch either from marketplace or Shared Image Gallery if the image has been enabled for trusted launch.
https://docs.microsoft.com/en-us/azure/virtual-machines/trusted-launch

Introduced two new parameters:
secure_boot_enabled
vtpm_enabled

### DIFF
--- a/builder/azure/common/template/template_builder.go
+++ b/builder/azure/common/template/template_builder.go
@@ -450,6 +450,38 @@ func (s *TemplateBuilder) SetBootDiagnostics(diagSTG string) error {
 	return nil
 }
 
+func (s *TemplateBuilder) SetSecurityProfile(secureBootEnabled bool, vtpmEnabled bool) error {
+	s.setVariable("apiVersion", "2020-12-01") // Required for Trusted Launch
+	resource, err := s.getResourceByType(resourceVirtualMachine)
+	if err != nil {
+		return err
+	}
+
+	if secureBootEnabled {
+		if resource.Properties.SecurityProfile == nil {
+			resource.Properties.SecurityProfile = &compute.SecurityProfile{}
+		}
+		if resource.Properties.SecurityProfile.UefiSettings == nil {
+			resource.Properties.SecurityProfile.UefiSettings = &compute.UefiSettings{}
+		}
+		resource.Properties.SecurityProfile.SecurityType = compute.SecurityTypesTrustedLaunch
+		resource.Properties.SecurityProfile.UefiSettings.SecureBootEnabled = to.BoolPtr(secureBootEnabled)
+	}
+
+	if vtpmEnabled {
+		if resource.Properties.SecurityProfile == nil {
+			resource.Properties.SecurityProfile = &compute.SecurityProfile{}
+		}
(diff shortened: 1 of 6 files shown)


builder/azure/common/template/template_builder.go
```diff
@@ -450,6 +450,22 @@ func (s *TemplateBuilder) SetBootDiagnostics(diagSTG string) error {
 	return nil
 }
 
+func (s *TemplateBuilder) SetSecurityProfile(secureBootEnabled bool, vtpmEnabled bool) error {
```
agrawalkshitij, line 453: is the destination image also being marked correctly as trusted launch?

maxilampert: A managed image has no specific setting. In case of the destination being a shared image gallery it won't set these properties that would need additional settings to be applied when publishing the image to to the gallery. But you can take an already created trusted launch image from a shared image gallery as source image

maxilampert: hmm it seems now that Trusted Launch VMs cannot be saved as managed images anymore, only direct publish to SIG is supported

builder/azure/common/template/template_builder.go
```diff
@@ -450,6 +450,22 @@ func (s *TemplateBuilder) SetBootDiagnostics(diagSTG string) error {
...
+	s.setVariable("apiVersion", "2020-12-01") // Required for Trusted Launch
+	resource, err := s.getResourceByType(resourceVirtualMachine)
+	if err != nil {
+		return err
+	}
+
+	resource.Properties.SecurityProfile = &compute.SecurityProfile{}
+	resource.Properties.SecurityProfile.UefiSettings = &compute.UefiSettings{}
```
JenGoldstrich, line 461: Hey @maxilampert this looks good to me but can you add a test to `template_builder_test.go` to capture these new fields, after that happy to merge this and get this functionality in, it's worth calling out that the vTPM info can't be captured in a deployed image, but for your use case for VMs it'll work

JenGoldstrich (changes requested):

builder/azure/common/template/template_builder.go
agrawalkshitij: Yes, there is a PR in progress. I think you will have to wait until that is done.

benz66: Hi There, is there any ideas as to when this enhancement will be generally available and useable in the wild?

paullisto-insight: hi there, +1 on this enhancement. excitedly waiting in anticipation

JenGoldstrich:
Hey @maxilampert, thanks again for submitting this contribution, amccullough84 has taken these changes and rebased on them on main with some test changes in https://github.com/hashicorp/packer-plugin-azure/pull/257, we're ironing out a validation step but that should be merged soon so i am closing this PR, I will make sure you are called out in the release notes for also working on this feature!

PR #223, opened 2022-08-12
feat: spot support

### LINKED ISSUE #37: Spot on Azure
_This issue was originally opened by @yellowhat as hashicorp/packer#8776. It was migrated here as a result of the [Packer plugin split](https://github.com/hashicorp/packer/issues/8610#issuecomment-770034737). The original body of the issue is below._

<hr>

Hi,
is it possible to use packer on Azure using spot instances?

#### Feature Description
Use spot instances on Azure

#### Use Case(s)
For devops purpose, spot instance can be a good way to reduce costs.

Thanks

mweibel:
Adds spot support for azure ARM. 

closes #37

### DIFF
--- a/builder/azure/arm/config.go
+++ b/builder/azure/arm/config.go
@@ -273,7 +273,7 @@ type Config struct {
 	// CLI example `az vm list-sizes --location westus`
 	VMSize string `mapstructure:"vm_size" required:"false"`
 
-	// If set will use spot instance type.
+	// If set use a spot instance during build; spot configuration settings only apply to the virtual machine launched by Packer and will not be persisted on the resulting image artifact.
 	Spot Spot `mapstructure:"spot" required:"false"`
 
 	// Specify the managed image resource group name where the result of the
--- a/builder/azure/arm/config.go
+++ b/builder/azure/arm/config.go
@@ -1,5 +1,5 @@
 //go:generate packer-sdc struct-markdown
-//go:generate packer-sdc mapstructure-to-hcl2 -type Config,SharedImageGallery,SharedImageGalleryDestination,PlanInformation
+//go:generate packer-sdc mapstructure-to-hcl2 -type Config,SharedImageGallery,SharedImageGalleryDestination,PlanInformation,Spot
 
 package arm
 
@@ -103,6 +103,13 @@ type SharedImageGalleryDestination struct {
 	// Specify a storage account type for the Shared Image Gallery Image Version.
 	// Defaults to `Standard_LRS`. Accepted values are `Standard_LRS`, `Standard_ZRS` and `Premium_LRS`
 	SigDestinationStorageAccountType string `mapstructure:"storage_account_type"`
+}
+
+type Spot struct {
+	// Specify eviction policy for spot instance: "Deallocate" or "Delete". If this is set, a spot instance will be used.
+	EvictionPolicy compute.VirtualMachineEvictionPolicyTypes `mapstructure:"eviction_policy"`
+	// How much should the VM cost maximally per hour. Specify -1 (or do not specify) to not evict based on price.
+	MaxPrice float32 `mapstructure:"max_price"`
 }
 
 type Config struct {
@@ -265,6 +272,9 @@ type Config struct {
 	//
 	// CLI example `az vm list-sizes --location westus`
 	VMSize string `mapstructure:"vm_size" required:"false"`
+
+	// If set will use spot instance type.
+	Spot Spot `mapstructure:"spot" required:"false"`
 
 	// Specify the managed image resource group name where the result of the
 	// Packer build will be saved. The resource group must already exist. If
@@ -1213,6 +1223,18 @@ func assertRequiredParametersSet(c *Config, errs *packersdk.MultiError) {
 		errs = packersdk.MultiErrorAppend(errs, fmt.Errorf("An os_type must be specified"))
 	} else {
 		errs = packersdk.MultiErrorAppend(errs, fmt.Errorf("The os_type %q is invalid", c.OSType))
+	}
+
+	/////////////////////////////////////////////
+	// Storage
+	if c.Spot.EvictionPolicy != "" {
+		if c.Spot.EvictionPolicy != compute.VirtualMachineEvictionPolicyTypesDelete && c.Spot.EvictionPolicy != compute.VirtualMachineEvictionPolicyTypesDeallocate {
+			errs = packersdk.MultiErrorAppend(errs, fmt.Errorf("The spot.eviction_policy %q is invalid, eviction_policy must be %q, %q, or unset", c.Spot.EvictionPolicy, compute.VirtualMachineEvictionPolicyTypesDelete, compute.VirtualMachineEvictionPolicyTypesDeallocate))
+		}
+	} else {
+		if c.Spot.MaxPrice != 0 {
+			errs = packersdk.MultiErrorAppend(errs, fmt.Errorf("Setting a spot.max_price without an spot.eviction_policy is invalid, eviction_policy must be %q or %q if max_price is set", compute.VirtualMachineEvictionPolicyTypesDelete, compute.VirtualMachineEvictionPolicyTypesDeallocate))
+		}
 	}
 
 	/////////////////////////////////////////////
--- a/builder/azure/arm/config_test.go
+++ b/builder/azure/arm/config_test.go
@@ -2439,3 +2439,97 @@ func TestConfigShouldAcceptValidCustomResourceBuildPrefix(t *testing.T) {
 		}
 	}
 }
+
+func TestConfigSpot(t *testing.T) {
+	config := map[string]interface{}{
+		"capture_container_name": "ignore",
+		"capture_name_prefix":    "ignore",
+		"image_publisher":        "ignore",
+		"image_offer":            "ignore",
+		"image_sku":              "ignore",
+		"location":               "ignore",
+		"storage_account":        "ignore",
+		"resource_group_name":    "ignore",
+		"subscription_id":        "ignore",
+		"os_type":                constants.Target_Linux,
+		"communicator":           "none",
+		"spot": map[string]interface{}{
+			"eviction_policy": "Deallocate",
+		},
+	}
+	var c Config
+	_, err := c.Prepare(config, getPackerConfiguration())
+	if err != nil {
+		t.Fatal("expected config to accept spot settings", err)
+	}
+}
+
+func TestConfigSpotInvalidEvictionPolicy(t *testing.T) {
+	config := map[string]interface{}{
+		"capture_container_name": "ignore",
+		"capture_name_prefix":    "ignore",
+		"image_publisher":        "ignore",
+		"image_offer":            "ignore",
+		"image_sku":              "ignore",
+		"location":               "ignore",
+		"storage_account":        "ignore",
+		"resource_group_name":    "ignore",
+		"subscription_id":        "ignore",
+		"os_type":                constants.Target_Linux,
+		"communicator":           "none",
+		"spot": map[string]interface{}{
+			"eviction_policy": "test",
+		},
+	}
+	var c Config
+	_, err := c.Prepare(config, getPackerConfiguration())
+	if err == nil {
+		t.Fatal("expected config to not accept spot settings", err)
+	}
+}
+
+func TestConfigSpotEmptyEvictionPolicy(t *testing.T) {
+	config := map[string]interface{}{
+		"capture_container_name": "ignore",
+		"capture_name_prefix":    "ignore",
+		"image_publisher":        "ignore",
+		"image_offer":            "ignore",
+		"image_sku":              "ignore",
+		"location":               "ignore",
+		"storage_account":        "ignore",
+		"resource_group_name":    "ignore",
+		"subscription_id":        "ignore",
+		"os_type":                constants.Target_Linux,
+		"communicator":           "none",
+		"spot":                   map[string]interface{}{},
+	}
+	var c Config
+	_, err := c.Prepare(config, getPackerConfiguration())
+	if err != nil {
+		t.Fatal("expected config to accept spot settings", err)
+	}
+}
+
+func TestConfigSpotEmptyEvictionPolicyMaxPriceSet(t *testing.T) {
+	config := map[string]interface{}{
+		"capture_container_name": "ignore",
+		"capture_name_prefix":    "ignore",
+		"image_publisher":        "ignore",
+		"image_offer":            "ignore",
+		"image_sku":              "ignore",
+		"location":               "ignore",
+		"storage_account":        "ignore",
+		"resource_group_name":    "ignore",
+		"subscription_id":        "ignore",
+		"os_type":                constants.Target_Linux,
+		"communicator":           "none",
+		"spot": map[string]interface{}{
+			"max_price": 100,
+		},
+	}
+	var c Config
+	_, err := c.Prepare(config, getPackerConfiguration())
+	if err == nil {
+		t.Fatal("expected config to not accept spot settings", err)
+	}
+}
--- a/builder/azure/common/template/template_builder.go
+++ b/builder/azure/common/template/template_builder.go
@@ -323,6 +323,22 @@ func (s *TemplateBuilder) SetAdditionalDisks(diskSizeGB []int32, dataDiskname st
 	return nil
 }
 
+func (s *TemplateBuilder) SetSpot(policy compute.VirtualMachineEvictionPolicyTypes, price float32) error {
+	resource, err := s.getResourceByType(resourceVirtualMachine)
+	if err != nil {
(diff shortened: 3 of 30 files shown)


builder/azure/arm/config_test.go
```diff
@@ -2390,6 +2390,7 @@ func TestConfigShouldRejectCustomDataAndCustomDataFile(t *testing.T) {
 	}
 }
 
+<<<<<<< HEAD
```
JenGoldstrich, line 2393: This PR has some artifacts from a merge, can you remove them?

builder/azure/common/template/template_builder.go
```diff
@@ -683,7 +699,7 @@ const BasicTemplate = `{
   },
   "variables": {
     "addressPrefix": "10.0.0.0/16",
-    "apiVersion": "2017-03-30",
+    "apiVersion": "2022-03-01",
```
JenGoldstrich, line 702: Does this PR require an API version bump this large to work?  If possible I would like to leave the API version the same and save API upgrading for a larger track of work we're planning here https://github.com/hashicorp/packer-plugin-azure/issues/243

builder/azure/arm/config.go
```diff
@@ -105,6 +105,13 @@ type SharedImageGalleryDestination struct {
 	SigDestinationStorageAccountType string `mapstructure:"storage_account_type"`
 }
 
+type Spot struct {
+	// Specify eviction policy for spot instance: "Deallocate" or "Delete". If this is set, a spot instance will be used.
+	EvictionPolicy string `mapstructure:"eviction_policy"`
```
JenGoldstrich, line 110: I might have missed it in this PR but can we input validation this string? I think this will currently allow you to send anything, I think we should throw an error in Packer if the user tries to send anything other than Deallocate or Delete

JenGoldstrich (changes requested):
Hey @mweibel thanks so much for your changes, I apologize for the delay in someone from the Packer team responding, I left some suggestions and noticed a place where there were some merge conflict artifacts, lemme know if you have any questions!

builder/azure/arm/config.go
mweibel: I adjusted the type and added validation. I'm not *that* familiar with the code base - can you please double check if that's fine or if I should use another approach (and maybe point me to an example I can use as a base in this case?). Thanks!

builder/azure/common/template/template_builder.go
mweibel: good question - I think I just updated to the latest version available but it does make sense to do this in a more general way. 

I'm not sure where to find the relevant information though. Looking at the [docs](https://learn.microsoft.com/en-us/azure/templates/microsoft.compute/2020-12-01/virtualmachines?pivots=deployment-language-arm-template) it only shows until 2020-12-01 and not earlier. Not sure where I could check that. Do you know?

JenGoldstrich: Hrm I'm not sure 🤔 , could you manually test your PR without the API bump, see if it fails, and if it doesn't work maybe let's just do the minimum API bump required, usually the error message will say what the required API verison is, if it doesn't I would recommend trying 2020-12-01 as the furthest bump we do for now

mweibel: I found [ARM schema 2019-03-01](https://github.com/Azure/azure-resource-manager-schemas/blob/025cb02ce0c1730099b4c76857099142e2cbf025/schemas/2019-03-01/Microsoft.Compute.json) to be the earliest which has the necessary attributes.

mweibel:
I updated the approval files, added also a separate test for spots and also did again a manual test which worked.

JenGoldstrich:
@mweibel I am thinking through a suggestion to the validation code but in the mean time I wanted to say that I'm not sure if I'm looking in the wrong place but when I build a Packer template with the below template

```
source "azure-arm" "this" {
    # General
    managed_image_resource_group_name = var.resource_group_name
    managed_image_name                = var.managed_image_name
    location                          = var.location
    vm_size                           = var.vm_size
    use_azure_cli_auth = true
    # # Image settings
    os_type         = var.image["os_type"]
    image_publisher = var.image["image_publisher"]
    image_offer     = var.image["image_offer"]
    image_sku       = var.image["image_sku"]

    # Communicator
    communicator   = "winrm"
    winrm_use_ssl  = true
    winrm_insecure = true
    winrm_timeout  = "5m"
    winrm_username = "packer"
    winrm_password = "Packer!!!567"

    spot {
        max_price = "100"
        eviction_policy = "Deallocate"
    }
}
```

I get the following image in Azure
```
[
  {
    "extendedLocation": null,
    "hyperVGeneration": "V1",
    "id": "/subscriptions/1f90521a-24f6-4758-ac3d-88d869fb0bf5/resourceGroups/jenna-test/providers/Microsoft.Compute/images/jenna-packer-test",
    "location": "westus",
    "name": "jenna-packer-test",
    "provisioningState": "Succeeded",
    "resourceGroup": "jenna-test",
    "sourceVirtualMachine": {
      "id": "/subscriptions/1f90521a-24f6-4758-ac3d-88d869fb0bf5/resourceGroups/pkr-Resource-Group-qucro9a52c/providers/Microsoft.Compute/virtualMachines/pkrvmqucro9a52c",
      "resourceGroup": "pkr-Resource-Group-qucro9a52c"
    },
    "storageProfile": {
      "dataDisks": [],
      "osDisk": {
        "blobUri": null,
        "caching": "ReadWrite",
        "diskEncryptionSet": null,
        "diskSizeGb": 127,
        "managedDisk": {
          "id": "/subscriptions/1f90521a-24f6-4758-ac3d-88d869fb0bf5/resourceGroups/pkr-Resource-Group-qucro9a52c/providers/Microsoft.Compute/disks/pkrosqucro9a52c",
          "resourceGroup": "pkr-Resource-Group-qucro9a52c"
        },
        "osState": "Generalized",
        "osType": "Windows",
        "snapshot": null,
        "storageAccountType": "Standard_LRS"
      },
      "zoneResilient": false
    },
    "tags": {},
    "type": "Microsoft.Compute/images"
  }
]

**Comment by JenGoldstrich** [comment]
Hey @mweibel I'm having trouble validating that spot is being set.

When trying to build with the following template 
```
source "azure-arm" "this" {
    # General
    managed_image_resource_group_name = var.resource_group_name
    managed_image_name                = var.managed_image_name
    location                          = var.location
    vm_size                           = var.vm_size
    use_azure_cli_auth = true
    # # Image settings
    os_type         = var.image["os_type"]
    image_publisher = var.image["image_publisher"]
    image_offer     = var.image["image_offer"]
    image_sku       = var.image["image_sku"]

    # Communicator
    communicator   = "winrm"
    winrm_use_ssl  = true
    winrm_insecure = true
    winrm_timeout  = "5m"
    winrm_username = "packer"
    winrm_password = "Packer!!!567"

    spot {
        max_price = "100"
        eviction_policy = "Deallocate"
    }
}
```
I don't see the spot being set when I try to create the VM, I also don't see the spot object in the properties block of the image or in the image json from az CLI, am I doing something wrong?

Screen Shot 2022-11-30 at 1 53 33 PM

When I set Deallocate to something incorrect the validation works so that's good, I wanna examine some other plugins and ask a colleague what they think about another edge case I'm thinking about involving max_price being set but eviction_policy is not that but didn't wanna wait on sending you this.

**Comment by mweibel** [comment]
hi @JenGoldstrich 

That's curious. I used the following template (based on `examples/ubuntu.json`):

```json
{
  "variables": {
    "client_id": "{{env `ARM_CLIENT_ID`}}",
    "client_secret": "{{env `ARM_CLIENT_SECRET`}}",
    "subscription_id": "{{env `ARM_SUBSCRIPTION_ID`}}",
    "tenant_id": "{{env `ARM_TENANT_ID`}}",
  },
  "builders": [{
    "type": "azure-arm",

    "client_id": "{{user `client_id`}}",
    "client_secret": "{{user `client_secret`}}",
    "subscription_id": "{{user `subscription_id`}}",
    "tenant_id": "{{user `subscription_id`}}",

    "os_type": "Linux",
    "image_publisher": "Canonical",
    "image_offer": "UbuntuServer",
    "image_sku": "16.04-LTS",

    "managed_image_resource_group_name": "packer",
    "managed_image_name": "TestUbuntuImage",

    "spot": {
      "max_price": "100",
      "eviction_policy": "Deallocate"
    },

    "location": "westeurope",
    "vm_size": "Standard_DS2_v2"
  }],
  "provisioners": [{
    "execute_command": "chmod +x {{ .Path }}; {{ .Vars }} sudo -E sh '{{ .Path }}'",
    "inline": [
      "apt-get update",
      "apt-get upgrade -y",

      "/usr/sbin/waagent -force -deprovision+user && export HISTSIZE=0 && sync"
    ],
    "inline_shebang": "/bin/sh -x",
    "type": "shell"
  }]
}
```

image

Also the ARM deployment template contains the following:

```json
[...]
        {
            "type": "Microsoft.Compute/virtualMachines",
            "apiVersion": "[variables('apiVersion')]",
            "name": "[parameters('vmName')]",
            "location": "[variables('location')]",
            "dependsOn": [
                "[concat('Microsoft.Network/networkInterfaces/', parameters('nicName'))]"
            ],
            "properties": {
                "billingProfile": {
                    "maxPrice": 100
                },
                "diagnosticsProfile": {
                    "bootDiagnostics": {
                        "enabled": false
                    }
                },
                "evictionPolicy": "Deallocate",
                "hardwareProfile": {
                    "vmSize": "[parameters('vmSize')]"
                },
[...]
```

Not sure what causes this difference. I doubt it's HCL vs JSON?

builder/azure/arm/config.go
```diff
@@ -1215,6 +1225,12 @@ func assertRequiredParametersSet(c *Config, errs *packersdk.MultiError) {
 		errs = packersdk.MultiErrorAppend(errs, fmt.Errorf("The os_type %q is invalid", c.OSType))
 	}
 
+	/////////////////////////////////////////////
+	// Storage
+	if c.Spot.EvictionPolicy != "" && c.Spot.EvictionPolicy != compute.VirtualMachineEvictionPolicyTypesDelete && c.Spot.EvictionPolicy != compute.VirtualMachineEvictionPolicyTypesDeallocate {
+		errs = packersdk.MultiErrorAppend(errs, fmt.Errorf("The spot.eviction_policy %q is not either empty, %q, or %q", c.Spot.EvictionPolicy, compute.VirtualMachineEvictionPolicyTypesDelete, compute.VirtualMachineEvictionPolicyTypesDeallocate))
+	}
+
```
JenGoldstrich, line 1233: I think we should capture the edge case where max price is set and not an eviction policy, this is currently not blocked and could be a bit confusing if the user didn't understand Spot since they could set a max_price and it would do nothing, what do you think?
```suggestion
	if c.Spot.EvictionPolicy != "" {
		if c.Spot.EvictionPolicy != compute.VirtualMachineEvictionPolicyTypesDelete && c.Spot.EvictionPolicy != compute.VirtualMachineEvictionPolicyTypesDeallocate {
			s = packersdk.MultiErrorAppend(errs, fmt.Errorf("The spot.eviction_policy %q is invalid, eviction_policy must be %q, %q, or unset", c.Spot.EvictionPolicy, compute.VirtualMachineEvictionPolicyTypesDelete, compute.VirtualMachineEvictionPolicyTypesDeallocate))
		}
	} else {
		if c.Spot.MaxPrice != 0 {
			s = packersdk.MultiErrorAppend(errs, fmt.Errorf("Setting a spot.max_price without an spot.eviction_policy is invalid, eviction_policy must be %q or %q if max_price is set", compute.VirtualMachineEvictionPolicyTypesDelete, compute.VirtualMachineEvictionPolicyTypesDeallocate))
		}
	}

```

JenGoldstrich:
I deployed the following template 
```
{
  "builders": [{
    "type": "azure-arm",

    "use_azure_cli_auth": true,
    "os_type": "Linux",
    "image_publisher": "Canonical",
    "image_offer": "UbuntuServer",
    "image_sku": "16.04-LTS",

    "managed_image_resource_group_name": "jenna-test",
    "managed_image_name": "TestUbuntuImage",

    "spot": {
      "max_price": "100",
      "eviction_policy": "Deallocate"
    },

    "location": "westeurope",
    "vm_size": "Standard_DS2_v2"
  }],
  "provisioners": [{
    "execute_command": "chmod +x {{ .Path }}; {{ .Vars }} sudo -E sh '{{ .Path }}'",
    "inline": [
      "apt-get update",
      "apt-get upgrade -y",

      "/usr/sbin/waagent -force -deprovision+user && export HISTSIZE=0 && sync"
    ],
    "inline_shebang": "/bin/sh -x",
    "type": "shell"
  }]
}


```
And then tried to create a VM with that image resulting in this VM
```
 {
            "type": "Microsoft.Compute/virtualMachines",
            "apiVersion": "2022-03-01",
            "name": "[parameters('virtualMachines_bweh_name')]",
            "location": "westeurope",
            "properties": {
                "hardwareProfile": {
                    "vmSize": "Standard_A1_v2"
                },
                "storageProfile": {
                    "imageReference": {
                        "id": "[parameters('images_TestUbuntuImage_externalid')]"
                    },
                    "osDisk": {
                        "osType": "Linux",
                        "name": "[concat(parameters('virtualMachines_bweh_name'), '')]",
                        "createOption": "FromImage",
                        "caching": "ReadWrite",
                        "managedDisk": {
                            "storageAccountType": "StandardSSD_LRS",
                            "id": "[parameters('')]"
                        },
                        "deleteOption": "Detach",
                        "diskSizeGB": 30
                    },
                    "dataDisks": []
                },
  ```
 
When I change Deallocate to an invalid string I get the Packer error introduced in this PR so I know I'm using the right version of the plugin, this definitely seems like a user error issue.  I also tried to just use the Azure CLI to rule out any UI issues with deploying the VM, I feel like this is more a user error and I will continue to investigate but please let me know if you think of anything this might be, gonna test on someone elses machine most likely to rule out environmental silliness

**Comment by mweibel** [comment]
I have no clue why it doesn't work for you but does for me ;) To be sure I also checked that I didn't have any local changes which didn't work or something like that.

I took your json and just adjusted the login to be via env variables instead of az cli and built it. It worked  well for me.
image

deployment template:

```
        },
        {
            "type": "Microsoft.Compute/virtualMachines",
            "apiVersion": "[variables('apiVersion')]",
            "name": "[parameters('vmName')]",
            "location": "[variables('location')]",
            "dependsOn": [
                "[concat('Microsoft.Network/networkInterfaces/', parameters('nicName'))]"
            ],
            "properties": {
                "billingProfile": {
                    "maxPrice": 100
                },
                "diagnosticsProfile": {
                    "bootDiagnostics": {
                        "enabled": false
                    }
                },
                "evictionPolicy": "Deallocate",
                "hardwareProfile": {
                    "vmSize": "[parameters('vmSize')]"
                },
```

what is interesting though, if I look at the JSON view of the VM I don't see the spot settings. The VM is spot though. 
Did you use the JSON view of the VM or the deployment json to check what's been set?

File: builder/azure/arm/config.go (hunk 4 above)
Comment (mweibel) [reply builder/azure/arm/config.go]: good idea 👍 thanks. Added + added a test for this case.

**Comment by mweibel** [comment]
Hi @JenGoldstrich

> My understanding originally was that the spot setting would persist to deployed VMs, however after trying to capture a VM with spot manually in the Azure UI without Packer we didn't see the captured Image support persisting the Spot information

Now I see where the confusion comes from :smile: 
Thanks for explaining your assumptions!

Yes the idea is to use spot machines for building the images. Runtime is a different story and needs to be configured when launching the VM with the built image. 

> If so why is it valuable to use Spot for these temporary VMs?

Economic and environmental reasons:
1. It's much cheaper to run spot VMs and some packer builds can take a while (especially windows builds are usually slower). While for one build this doesn't make that much difference, over time this leads to significant cheaper builds. 
2. Using spot VMs means improving the usage of the host machines in the respective datacenters. Idle hosts just use energy and with spot VMs we can contribute to better host computer usage and therefore help our environment. 

Maybe people in #37 have additional reasons, these are mine. 

Does that make sense?

File: docs-partials/builder/azure/arm/Config-not-required.mdx [hunk 5]
```diff
@@ -111,6 +111,8 @@
   
   CLI example `az vm list-sizes --location westus`
 
+- `spot` (Spot) - If set will use spot instance type.
```
JenGoldstrich, line 114:
```suggestion
- `spot` (Spot) - If set use a spot instance type at runtime for the build; spot configuration settings only apply to the virtual machine launched by Packer and will not be persisted on the resulting image artifact.
```

JenGoldstrich:
That does make sense @mweibel, I'm a bit concerned users will make the same incorrect assumption, I made a docs change to hopefully make this feature a bit clearer, what do you think?

mweibel:
@JenGoldstrich that does make sense, thanks. I added the docs with a slight modification which seemed easier to understand for me. Let me know if you disagree. 

Thanks!

PR #225, opened 2022-08-12
Update max SIG replica count from 10 to 100

arsenvlad:
Update the maximum image replica from the previous value of 10 to the new value of 100.

Azure Compute Gallery recently updated the limits of image replica counts from 10 to 100, but 50 replicas should be sufficient for most use cases.

https://docs.microsoft.com/en-us/azure/virtual-machines/azure-compute-gallery#limits

image


JenGoldstrich (approved):
LGTM thanks for that link to the Microsoft docs that's very appreciated! @arsenvlad

PR #228, opened 2022-08-18
Fix err check to show error correctly

matt-richardson:
While working on some other changes, I found that the error checking for "did the deployment succeed" was incorrect - it was showing the error message if there was no error, rather than if there was an error.

(I think I can get away with not including tests for this, given it's just a UI change?)


nywilken (approved):
Great catch! LGTM

PR #252, opened 2022-12-08
go.mod: bump go version from 1.17 to 1.18

lbajolet-hashicorp:


nywilken (approved):
Approving - but I would recommend waiting for @JenGoldstrich okay to avoid any possible go.mod conflicts on PRs she is currently reviewing.

PR #255, opened 2022-12-20
[CR254] Set "image_version" to "default" in all sample config files

nkadel:
Feature request 254 requests adding the "image_version" set to the default "latest" value consistently in every sample file. Some have it, some don't, let's be consistent if we can.

### DIFF
--- a/builder/azure/arm/builder_acc_test.go
+++ b/builder/azure/arm/builder_acc_test.go
@@ -380,6 +380,7 @@ func testBuilderUserDataLinux(userdata string) string {
 	  "image_publisher": "Canonical",
 	  "image_offer": "UbuntuServer",
 	  "image_sku": "16.04-LTS",
+	  "image_version": "latest",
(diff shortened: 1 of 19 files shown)


JenGoldstrich (changes requested):
Hey @nkadel, thanks for making this contribution.

I think it would be better if this change only applied to the templates in the example directory and the docs/builders directory, the other templates you've changed are not meant to be examples, and are instead used as part of automated testing, and it makes the change a bit more disruptive, can we only apply it to templates we don't use in automated testing?

JenGoldstrich:
Closing due to inactivity, but we should update these examples soon as they are very out of date I agree, I would just prefer to focus on adding new examples that use HCL, and plugin features added recently.

nkadel:
I don't see why these can't be set explicitly everywhere, to allow the pipelines access to override them as desired. Putting in the "version" of "latest" simply makes clear that there is, indeed, an additional variable and its default value.

Issue #221, opened 2022-08-04
Support Trusted Launch VMs as source image


maxilampert:
#### Community Note

Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request.
Please do not leave "+1" or "me too" comments, they generate extra noise for issue followers and do not help prioritize the request.
If you are interested in working on this issue or have submitted a pull request, please leave a comment.

#### Description

Allow packer to create VMs that are using the trusted launch feature for SecureBoot and VTPM. When used by with a shared image gallery there is only and on/off switch for this feature meaning that when using a trusted lanuch image from the gallery it is currently not possible to create a new image based on that

#### Use Case(s)

Using chained builds we today have a shared image gallery that contains a base image and on top of that additional images get created but when the base image has been pushed as a trusted launch VM it is not possible for packer to use it as a source for the next image

#### Potential configuration

```
```

#### Potential References
https://docs.microsoft.com/en-us/azure/virtual-machines/trusted-launch

PR #257, opened 2023-01-08
Enable SecurityProfile (TrustedLaunch) for Virtual Machines

amccullough84:
This PR extends the existing PR https://github.com/hashicorp/packer-plugin-azure/pull/220 but adds the requested change from @JenGoldstrich to add tests for the new parameters.

The previous PR has been idle for several months and as this is a popular request I'm hoping this can be progressed.

As per the previous discussion - TrustedLaunch only applies to Virtual Machines, not images, however an image to be used with for a Trusted Launch enabled VM must be created using a builder VM with trusted launch enabled. It cannot be enabled after initial deployment.

Description of change from existing PR (220):

Sets the SecurityProfile if either Secure Boot or the VTPM option has been set in the packer configuration. This allows the deployment of VMs using Trusted Launch either from marketplace or Shared Image Gallery if the image has been enabled for trusted launch.

Introduced two new parameters:
secure_boot_enabled
vtpm_enabled

Closes https://github.com/hashicorp/packer-plugin-azure/issues/221

### DIFF
--- a/builder/azure/arm/config.go
+++ b/builder/azure/arm/config.go
@@ -1068,6 +1068,10 @@ func assertRequiredParametersSet(c *Config, errs *packersdk.MultiError) {
 		errs = packersdk.MultiErrorAppend(errs, fmt.Errorf("A managed image must be created from a managed image, it cannot be created from a VHD."))
 	}
 
+	if (c.SecureBootEnabled || c.VTpmEnabled) && isCustomManagedImage {
+		errs = packersdk.MultiErrorAppend(errs, fmt.Errorf("A managed image can not set SecureBoot or VTpm, these features are only supported when directly publishing to a Shared Image Gallery"))
+	}
+
 	if (c.CaptureContainerName != "" || c.CaptureNamePrefix != "" || c.ManagedImageName != "") && c.DiskEncryptionSetId != "" {
 		errs = packersdk.MultiErrorAppend(errs, fmt.Errorf("Setting a disk encryption set ID is not allowed when building a VHD or creating a Managed Image, only when publishing directly to Shared Image Gallery"))
 	}
--- a/builder/azure/arm/config.go
+++ b/builder/azure/arm/config.go
@@ -1068,7 +1068,7 @@ func assertRequiredParametersSet(c *Config, errs *packersdk.MultiError) {
 		errs = packersdk.MultiErrorAppend(errs, fmt.Errorf("A managed image must be created from a managed image, it cannot be created from a VHD."))
 	}
 
-	if (c.SecureBootEnabled || c.VTpmEnabled) && (c.ManagedImageName != "" && c.ManagedImageResourceGroupName != "") {
+	if (c.SecureBootEnabled || c.VTpmEnabled) && (c.ManagedImageName != "" || c.ManagedImageResourceGroupName != "") {
 		errs = packersdk.MultiErrorAppend(errs, fmt.Errorf("A managed image (managed_image_name, managed_image_resource_group_name) can not set SecureBoot or VTpm, these features are only supported when directly publishing to a Shared Image Gallery"))
 	}
 
--- a/builder/azure/arm/config.go
+++ b/builder/azure/arm/config.go
@@ -1068,8 +1068,8 @@ func assertRequiredParametersSet(c *Config, errs *packersdk.MultiError) {
 		errs = packersdk.MultiErrorAppend(errs, fmt.Errorf("A managed image must be created from a managed image, it cannot be created from a VHD."))
 	}
 
-	if (c.SecureBootEnabled || c.VTpmEnabled) && isCustomManagedImage {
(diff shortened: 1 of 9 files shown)


JenGoldstrich:
Hey @amccullough84 thanks for working on that,

This PR I think is actually blocked because of https://github.com/hashicorp/packer-plugin-azure/pull/220#discussion_r973801422, where we mention that secure launch VMs can't be saved as managed images, only deployed to SIGs, so I believe we need to finish merging https://github.com/hashicorp/packer-plugin-azure/pull/242, another open PR I am currently working on getting merged before we can get this working fully, I could be misunderstanding this feature though.  I will come back to this PR once that PR is merged.

JenGoldstrich:
@amccullough84 can you rebase your PR on main please so that I can test that these flags work when publishing directly to SIG without a managed image?

amccullough84:
@JenGoldstrich rebased on main now 👍

JenGoldstrich:
I've tried to build an image with this PR's changes am running into some failures.

Specifically 
```
Build 'azure-arm.autogenerated_1' errored after 13 minutes 12 seconds: Code="Conflict" Message="The source '/subscriptions/(subscription)/resourceGroups/pkr-Resource-Group-pl7hj39oj0/providers/Microsoft.Compute/virtualMachines/pkrvmpl7hj39oj0' has security type 'TrustedLaunch' and cannot be used as a source for an image definition with SecurityType feature set to 'None'."
```

With the following build file:
```
source "azure-arm" "autogenerated_1" {
  communicator                      = "winrm"
  image_offer                       = "Windows-10"
  image_publisher                   = "MicrosoftWindowsDesktop"
  image_sku                         = "19h2-pro-g2"
  location                          = "West US"
  use_azure_cli_auth = "true"
  os_type                           = "Windows"
  vm_size                           = "Standard_DS2_v2"
  winrm_insecure                    = "true"
  winrm_timeout                     = "3m"
  winrm_use_ssl                     = "true"
  winrm_username                    = "packer"
  secure_boot_enabled = true
  vtpm_enabled = true
    resource_group = "jenna"
    gallery_name = "trustedwindows2016"
    image_name = "image"
    image_version = var.version
    storage_account_type = "Standard_LRS"
  }
}
```

I don't think we support setting a security type, but I've tried with multiple different trusted launch VMs and have gotten this error, how are you getting around it for your builds? Is there a way to tell which images would be supported, or is there something I need to change on the Azure side?

amccullough84:
Does the image definition in the ACG you're distributing the image to have the security type set to trusted launch? 

You'd set that when creating the image definition and the error suggests there's a mismatch. If its enabled on the source image it must be enabled in the destination image definition.

JenGoldstrich (commented):
Oops! That makes sense, I fixed that and was able to push to my compute gallery, nice!

When I try to build this while also publishing to a managed image, autorest panics and crashes and I get an Azure error above that about not being able to publish trusted launch images to managed images as expected, but I think it would be good to add a check in the arm config.go in the assertRequiredParametersSet function right after the 
```
if isImageUrl && c.ManagedImageResourceGroupName != "" 
```
error check that is something like 
```
if (c.SecureBootEnabled || c.VTpmEnabled) && isCustomManagedImage) {
		errs = packersdk.MultiErrorAppend(errs, fmt.Errorf("A managed image can not set SecureBoot or VTpm, these features are only supported when directly publishing to a Shared Image Gallery"))
}
```

What do you think @amccullough84?

amccullough84:
Hi @JenGoldstrich, sorry for the delay - I've added that check now and seems to catch that conflict now when I run it locally.

builder/azure/arm/config.go
```diff
@@ -1063,6 +1068,10 @@ func assertRequiredParametersSet(c *Config, errs *packersdk.MultiError) {
 		errs = packersdk.MultiErrorAppend(errs, fmt.Errorf("A managed image must be created from a managed image, it cannot be created from a VHD."))
 	}
 
+	if (c.SecureBootEnabled || c.VTpmEnabled) && (c.ManagedImageName != "" && c.ManagedImageResourceGroupName != "") {
```
JenGoldstrich, line 1071: @amccullough84 changed this flag, and added tests for it

JenGoldstrich (commented):
@amccullough84 hey no worries friend, I actually messed up and told you the wrong config field, this should block the creation of managed_images, not using custom_managed_image as a base image field, Trusted Launch is only supported when direct publishing to a compute gallery/shared image gallery as mentioned, I've made a commit https://github.com/hashicorp/packer-plugin-azure/pull/257/commits/7e80049662f8c9bc463d8c45a5c90db4c0ec6080 to your fork directly to make that change, and to add unit tests for that suggested change.

Can you take a look and give me a thumbs up if this still works for your use case and that I haven't broken this for you, and then this should be good to merge

amccullough84:
Hi @JenGoldstrich, looks good to me!

JenGoldstrich (approved):

Issue #181, opened 2022-02-02
Support for Azure customer managed keys (CMK) on OS disk


hc-github-team-packer:
_This issue was originally opened by @mickeyoc76 in https://github.com/hashicorp/packer/issues/11545 and has been migrated to this repository. The original issue description is below._

---

#### Overview of the Issue

Our company Azure policy dictates that all virtual machines should be created with a disk encryption set with customer managed keys. (CMK)
My initial research cannot find any reference to this in the packer documentation, can you confirm if this is something configurable in the packer.json ?  Or something that is supported?

#### Reproduction Steps

Running a packer build fails during machine creation with policy violation

### Packer version

1.6.5

###

PR #258, opened 2023-01-09
Enable DIsk Encryption Set support for build VMs

JenGoldstrich:
WIP

probably closes https://github.com/hashicorp/packer-plugin-azure/issues/181
probably closes https://github.com/hashicorp/packer-plugin-azure/issues/47


JenGoldstrich:
This branch got messed up because of a weird local state of git remotes while testing PRs, moving to another PR

Issue #215, opened 2022-07-06
Azure-arm Packer doesn't clean up additional disk


jkasnar2410:
#### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* Please do not leave "+1" or other comments that do not add relevant new information or questions, they generate extra noise for issue followers and do not help prioritize the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

#### Overview of the Issue
There seems to be an issue with Packer not deleting an additional disk that was attached to the build machine, regardless if the build fails or not. **The issue seems to only be happening when a custom resource group is used for the build.** The OS disk gets deleted normally, but the additional one is always left there in an unattached state. There is no error during clean up, so it seems to be that it doesn't even try to delete the additional disk.
If a build is run when no resource group is specified (packer created a new one), the disk gets normally deleted.
This seems to be the same issue already reported that was apparently fixed 2 years ago:
https://github.com/hashicorp/packer/issues/10268
https://github.com/hashicorp/packer/pull/10713

#### Reproduction Steps

1. Write a simple Packer template for azure-arm that uses existing resource group
2. Wait for the build to finish or fail (it's irelevant )
3. Check Azure portal and see the disk still there in an unattached state

### Packer version

1.8.2 (latest ATM)

### Simplified Packer Template

```
source "azure-arm" "appliance-azure-image-2disks" {
  image_offer                            = "rockylinux"
  image_publisher                        = "erockyenterprisesoftwarefoundationinc1653071250513"
  image_sku                              = "free"
  managed_image_name                     = "test-disk-build"
  managed_image_resource_group_name      = "<< my_resource_group >>"
  os_type                                = "Linux"
  subscription_id                        = "<< my_subscription_id"
  tenant_id                              = "<< my_tenant_id >>"
  vm_size                                = "Standard_D4as_v4"
  os_disk_size_gb                        = "50"
  build_resource_group_name              = "<< my_resource_group >>"
  virtual_network_name                   = "<< my_virtual_network >>"
  virtual_network_subnet_name            = "<< my_subnet >>"
  virtual_network_resource_group_name    = "<< my_resource_group >>"
  private_virtual_network_with_public_ip = "true"
  skip_create_image                      = "false"
  custom_resource_build_prefix           = "ado"
[...]

PR #259, opened 2023-01-09
Support encrypting disks for Shared Image Galleries, Cleanup Additional Disks

JenGoldstrich:
This PR adds disk encryption set ID support for encrypting OS and data disks when publishing to a compute gallery.

This PR also fixes a bug where previously additional disks were not cleaned up, instead we'd only delete them when deleting the resource group, so if users provided their own build resource group additional disks would be left behind unattached 

I only implemented this for SIG as whenever I tried for managed images I ran into API errors saying I could not set an encryption key on the CreateImage endpoint when using a source VM, the Azure Portal also doesn't seem to allow this, and I could not find a resource template example that didn't use compute galleries when publishing encrypted images, I could not find anything concrete in the Azure docs but I believe Azure might only support encrypting disks this way when publishing directly to SIG, similar to ARM64 Windows VMs

https://learn.microsoft.com/en-us/azure/virtual-machines/image-version-encryption

closes https://github.com/hashicorp/packer-plugin-azure/issues/181 
closes https://github.com/hashicorp/packer-plugin-azure/issues/47
closes https://github.com/hashicorp/packer-plugin-azure/issues/215


builder/azure/arm/step_delete_additional_disks.go
```diff
@@ -67,19 +67,13 @@ func (s *StepDeleteAdditionalDisk) Run(ctx context.Context, state multistep.Stat
...
 	var isManagedDisk = state.Get(constants.ArmIsManagedImage).(bool)
-	var isExistingResourceGroup = state.Get(constants.ArmIsExistingResourceGroup).(bool)
 	var resourceGroupName = state.Get(constants.ArmResourceGroupName).(string)
 
 	if dataDisks == nil {
 		s.say(" -> No Additional Disks specified")
 		return multistep.ActionContinue
 	}
 
-	if isManagedDisk && !isExistingResourceGroup {
```
JenGoldstrich, line 78: @nywilken this check has been in the code since this logic was implemented back in 2018, I think we don't properly clean up additional disks for managed images ever as shown in: https://github.com/hashicorp/packer-plugin-azure/issues/215, I don't think we should keep these additional VM disks regardless of if we're in an existing resource group we're using or not for the data disks but I'm not 100% certain my logic, what do you think?

builder/azure/common/template/template_builder.go
```diff
@@ -719,7 +745,7 @@ const BasicTemplate = `{
   },
   "variables": {
     "addressPrefix": "10.0.0.0/16",
-    "apiVersion": "2019-03-01",
+    "apiVersion": "2021-11-01",
```
JenGoldstrich, line 748: Had to bump API version for disk encryption set support

builder/azure/arm/step_delete_additional_disks.go
JenGoldstrich: OH Actually this step isn't even used!!! That's why the additional disks aren't deleted, we rely on the cleanup of builder/azure/arm/step_create_resource_group.go to delete it, so if we don't delete the resource group we don't delete the additional disks!

JenGoldstrich:
So this gets past the first encryption check currently (when a VM is created), now I'm running into failures with SIG, and i imagine managed images will also be impacted where they're failing to publish, I need to figure out which fields to add the encryption to for these two resources and that should be resolved though

builder/azure/arm/template_factory.go
```diff
@@ -156,9 +156,16 @@ func GetVirtualMachineDeployment(config *Config) (*resources.Deployment, error)
...
+		err = builder.SetDiskEncryptionSetID(config.BuildDiskEncryptionSetId)
+		if err != nil {
+			return nil, err
+		}
+	}
+
 	if len(config.AdditionalDiskSize) > 0 {
-		isManaged := config.CustomManagedImageName != "" || (config.ManagedImageName != "" && config.ImagePublisher != "") || config.SharedGallery.Subscription != ""
-		err = builder.SetAdditionalDisks(config.AdditionalDiskSize, config.tmpDataDiskName, isManaged, config.diskCachingType)
+		isLegacyVHD := config.CustomManagedImageName == "" && config.ManagedImageName == "" && config.ImagePublisher == "" && config.SharedGalleryDestination.SigDestinationGalleryName == ""
```
JenGoldstrich, line 167: "isManaged" is a bit confusing here, since we want to set the ManagedDisk properties even if just directly publishing to SIG, this was setup to check the SIG subscription, but it should be set to the SIG Destination gallery name, as you don't have to set a subscription if you just inherit it from higher up in the template

docs-partials/builder/azure/arm/Config-not-required.mdx
```diff
@@ -239,13 +241,6 @@
   See [documentation](https://docs.microsoft.com/en-us/azure/virtual-machines/extensions/custom-script-windows)
   to learn more.
 
-  The following example executes the contents of the file specified by `user_data_file`:
```
JenGoldstrich, line 242: This got removed cause it was put it in a generated doc file rather than the source document, I will fix this before releasing most likely in a seperate PR

builder/azure/common/template/template.go
```diff
@@ -52,19 +60,19 @@ type OSDiskUnion struct {
...
-	ManagedDisk  *compute.ManagedDiskParameters `json:"managedDisk,omitempty"`
+	Lun          *int                          `json:"lun,omitempty"`
+	BlobURI      *string                       `json:"blobUri,omitempty"`
+	Name         *string                       `json:"name,omitempty"`
+	Vhd          *compute.VirtualHardDisk      `json:"vhd,omitempty"`
+	Image        *compute.VirtualHardDisk      `json:"image,omitempty"`
+	Caching      compute.CachingTypes          `json:"caching,omitempty"`
+	CreateOption compute.DiskCreateOptionTypes `json:"createOption,omitempty"`
+	DiskSizeGB   *int32                        `json:"diskSizeGB,omitempty"`
+	ManagedDisk  *ManagedDisk                  `json:"managedDisk,omitempty"`
```
lbajolet-hashicorp, line 75: Am I right to assume this will be a breaking change? Is there a way we can avoid this potentially?

builder/azure/arm/step_publish_to_shared_image_gallery.go
```diff
@@ -172,18 +182,23 @@ func (s *StepPublishToSharedImageGallery) Run(ctx context.Context, stateBag mult
 		miSigReplicaCount = constants.SharedImageGalleryImageVersionDefaultMaxReplicaCount
 	}
 
+	buildDiskEncryptionSetId := stateBag.Get(constants.ArmBuildDiskEncryptionSetId).(string)
```
lbajolet-hashicorp, line 185: Is this attribute always set? Otherwise I'm slightly worried this will crash on the type cast

builder/azure/arm/config.go
```diff
@@ -343,7 +343,11 @@ type Config struct {
 	BuildKeyVaultName string `mapstructure:"build_key_vault_name"`
 	// Specify the KeyVault SKU to create during the build. Valid values are
 	// standard or premium. The default value is standard.
-	BuildKeyVaultSKU           string `mapstructure:"build_key_vault_sku"`
+	BuildKeyVaultSKU string `mapstructure:"build_key_vault_sku"`
+
+	// Specify the Disk Encryption Set ID to use to encrypt the OS and data disks created with the VM during the build
+	DiskEncryptionSetId string `mapstructure:"disk_encryption_set_id"`
```
lbajolet-hashicorp, line 349: I'm not familiar with Azure, but what are the accepted values here? Just looking at the attribute I'm not sure what is expected as far as IDs go for disk encryption

JenGoldstrich: Good question!

Azure supports three different types of disk encryption, platform managed keys (the default), customer managed keys (CMKs), or a combination of both keys,
Screen Shot 2023-01-25 at 1 35 50 PM

These CMKs are stored in Disk Encryption Sets, which have IDs,

These IDs are structured like

`/subscriptions/(subscription-uuid)/resourceGroups/jenna/providers/Microsoft.Compute/diskEncryptionSets/name`, 

Screen Shot 2023-01-25 at 1 27 44 PM

It felt simpler to just pass all this in as a string, but we could also take this input in an object that takes in the subscription, resource group, and encryption set of the Disk encryption set, where we assume the subscription and the resource group based on other fields, do you think that would make more sense?

builder/azure/common/template/template.go
JenGoldstrich: I don't believe this is a breaking change, this object is our definition of the Azure VM template object, we define a structure of that object that we use before marshalling it into a JSON that is used to create the VM, adding this field just adds the structure to set the encryption ID.  If you don't set the encryption ID the managed_disk field the resulting JSON should be the same as it was before this change, since we were already defining the managedDIsk json object

lbajolet-hashicorp: Oh yeah you're right, for some reason I assumed this was a change to the config for the builder, but it looks like it's completely internal, never mind!

builder/azure/arm/step_publish_to_shared_image_gallery.go
JenGoldstrich: Oops! It's not, and it does crash, good catch! I've addressed this in the newest commit

builder/azure/arm/config.go
lbajolet-hashicorp: Eh, it's mostly my lack of knowledge about Azure expressing itself in my question :)
Both alternatives are fine imho, if the resource ID is clearly exposed in the console, I'd think it's easier to specify rather than specifying the rest. Maybe to help users understand what is expected we could point them to some Azure docs through a link here?

docs-partials/builder/azure/arm/Config-not-required.mdx
```diff
@@ -234,17 +238,10 @@
...
-  See [documentation](https://docs.microsoft.com/en-us/azure/virtual-machines/extensions/custom-script-windows)
-  to learn more.
-
-  The following example executes the contents of the file specified by `user_data_file`:
-
-  ```hcl2
-  custom_script   = "powershell -ExecutionPolicy Unrestricted -NoProfile -NonInteractive -Command \"$userData = (Invoke-RestMethod -Headers @{Metadata=$true} -Method GET -Uri http://169.254.169.254/metadata/instance/compute/userData?api-version=2021-01-01$([char]38)format=text); $contents = [System.Text.Encoding]::UTF8.GetString([Convert]::FromBase64String($userData)); set-content -path c:\\Windows\\Temp\\userdata.ps1 -value $contents; . c:\\Windows\\Temp\\userdata.ps1;\""
-  user_data_file  = "./scripts/userdata.ps1"
-  ```
+- `custom_script` (string) - ```hcl2
```
nywilken, line 241: This seems to be missing the documentation for the custom_script attribute.

builder/azure/common/template/template.go
```diff
@@ -52,19 +60,19 @@ type OSDiskUnion struct {
 	Caching      compute.CachingTypes              `json:"caching,omitempty"`
 	CreateOption compute.DiskCreateOptionTypes     `json:"createOption,omitempty"`
 	DiskSizeGB   *int32                            `json:"diskSizeGB,omitempty"`
-	ManagedDisk  *compute.ManagedDiskParameters    `json:"managedDisk,omitempty"`
+	ManagedDisk  *ManagedDisk                      `json:"managedDisk,omitempty"`
```
nywilken, line 63: This is because of the API version change, correct?

nywilken (approved):
I left two small comments but this otherwise looks good to me. Have you ran the acceptance tests against this PR?

builder/azure/common/template/template.go
JenGoldstrich: for ManagedDiskParameters, I was having trouble setting the DiskEncryptionID so I just defined more of the object in our template, it's the same json that the compute package uses

JenGoldstrich:
@nywilken I messaged you off of GitHub but just for context for those watching here, acceptance tests are passing (minus a DTL flake that is not related), I wanna clean up the docs, maybe write some coverage tests, and do some double checking on data disks and then I'll merge.

JenGoldstrich:
@nywilken DTL acceptance tests are passing after setting the polling timeout to 25 minutes!

builder/azure/arm/config.go
JenGoldstrich: I added some clarification about where to find that resource ID in the docs

PR #293, opened 2023-04-13
docs(spot) ensure a block is used by adding example and including missing attributes partial

dduportal:
Hello there 👋 

While trying to enable the `spot` mode introduced in https://github.com/hashicorp/packer-plugin-azure/pull/223, 
we struggle to get that the `spot` atrribute is in fact a block (ref.  https://github.com/jenkins-infra/packer-images/pull/598).

This PR proposes an update of the documentation with the following enhancements:

- Add an example for JSON and HCL for the `spot` block
- Add a "Spot" section which include the autogenerated partial block to ensure all attributes of the `spot`  block are defined

### DIFF
--- a/docs-partials/builder/azure/arm/Config-not-required.mdx
+++ b/docs-partials/builder/azure/arm/Config-not-required.mdx
@@ -108,10 +108,27 @@
   VM from your VHD. See
   [pricing](https://azure.microsoft.com/en-us/pricing/details/virtual-machines/)
   information. Defaults to `Standard_A1`.
-  
+
   CLI example `az vm list-sizes --location westus`
 
 - `spot` (Spot) - If set use a spot instance during build; spot configuration settings only apply to the virtual machine launched by Packer and will not be persisted on the resulting image artifact.
+
+Following is an example:
+
+  In JSON
+  ```json
+  "spot": {
+      "eviction_policy": "Delete",
+      "max_price": "0.4",
+  }
+  ```
+  In HCL2
+  ```hcl
+  spot {
+      eviction_policy = "Delete"
+      max_price = "0.4"
+  }
+  ```
 
 - `managed_image_resource_group_name` (string) - Specify the managed image resource group name where the result of the
   Packer build will be saved. The resource group must already exist. If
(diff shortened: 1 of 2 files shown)


docs-partials/builder/azure/arm/Config-not-required.mdx
```diff
@@ -108,11 +108,28 @@
   VM from your VHD. See
   [pricing](https://azure.microsoft.com/en-us/pricing/details/virtual-machines/)
   information. Defaults to `Standard_A1`.
-  
+
```
dduportal, line 111:
```suggestion
  
```

docs-partials/builder/azure/arm/Config-not-required.mdx
```diff
@@ -113,6 +113,23 @@
 
 - `spot` (Spot) - If set use a spot instance during build; spot configuration settings only apply to the virtual machine launched by Packer and will not be persisted on the resulting image artifact.
 
+Following is an example:
+
+  In JSON
```
JenGoldstrich, line 118: Just just a heads up this is a generated file (see line 1) that is generated based on the comments in `builder/azure/arm/config.go` https://github.com/hashicorp/packer-plugin-azure/blob/353eea159082d35fcbaebaafb6708b57aee2fa2a/builder/azure/arm/config.go#L279 here

Here is an example of an HCL/JSON example in this format https://github.com/hashicorp/packer-plugin-azure/blob/353eea159082d35fcbaebaafb6708b57aee2fa2a/builder/azure/arm/config.go#L173-L197

Than you run `make generate` which will generate the docs-partials.  If I run `make generate` right now on your branch, it deletes your additional changes

JenGoldstrich (changes requested):
Thanks for this contribution! We definitely should add examples for these blocks I agree I added a note about the docs-partials being generated files, if you can address that I'd be happy to merge this

docs-partials/builder/azure/arm/Config-not-required.mdx
dduportal: Thanks for the review and clear explanation! I've added a commit with the source update (and the result of `make generate`).

Is that ok ?

JenGoldstrich (approved):
LGTM! Thanks for the contributions!

Issue #190, opened 2022-04-01
Failed to find temporary OS disk on VM.  Please delete manually.


xpillons:
#### Overview of the Issue

Packer is failing when building a Windows VM in an existing RG with the following error message 
`Failed to find temporary OS disk on VM.  Please delete manually.`
This makes our pipeline to failed.

#### Reproduction Steps

Try to build an image with the below config file.

### Plugin and Packer version

1.8.0

### Simplified Packer Buildfile

```
{
    "builders": [
        {
            "type": "azure-arm",
            "use_azure_cli_auth": "{{user `var_use_azure_cli_auth`}}",
            "image_publisher": "MicrosoftWindowsDesktop",
            "image_offer": "Windows-10",
            "image_sku": "21h1-pron",
            "image_version": "latest",
            "managed_image_resource_group_name": "{{user `var_resource_group`}}",
            "managed_image_name": "{{user `var_image`}}",
            "os_type": "Windows",
            "vm_size": "Standard_d8s_v3",
            "managed_image_storage_account_type": "Premium_LRS",

            "communicator": "winrm",
            "winrm_use_ssl": true,
            "winrm_insecure": true,
            "winrm_timeout": "5m",
            "winrm_username": "packer",

            "build_resource_group_name": "{{user `var_resource_group`}}",
            "cloud_environment_name": "{{user `var_cloud_env`}}"
        }
    ],
    "provisioners": [{
        "type": "powershell",
        "scripts": [
          "./scripts/windows/wait-for-agent.ps1",
          "./scripts/windows/install-choco.ps1",
          "./scripts/windows/optimize.ps1",
          "./scripts/windows/sysprep.ps1"
        ]
      }]
}

```

### Operating system and Environment details

Running packer from WSL2 - Ubuntu 20.04 

### Log Fragments and crash.log files

```
2022-04-01T18:07:30+02:00: ==> azure-arm: Waiting for deletion of all resources...
2022-04-01T18:07:34+02:00: ==> azure-arm: Failed to find temporary OS disk on VM.  Please delete manually.
==> azure-arm: 
==> azure-arm: VM Name: pkrvmx6hsnxrcvj
==> azure-arm: Error: %!s(<nil>)
2022-04-01T18:07:34+02:00: ==> azure-arm: Removing the created Deployment object: 'kvpkrdpx6hsnxrcvj'
2022-04-01T18:07:51+02:00: ==> azure-arm: 
[...]

PR #296, opened 2023-04-21
Deploy template error handling cleanup

JenGoldstrich:
Currently `step_deploy_template` is used for deploying KeyVaults, and Virtual Machines, however on CleanUp of this step we try and find a VM either way, and then return an "error" if we don't find the VM, this leads to every build falsely showing that we failed to remove the temporary disk

This PR breaks this logic up, and only tries to search for a VM disk when we're actually in a VM deploy

Partially addresses https://github.com/hashicorp/packer-plugin-azure/issues/190

### DIFF
--- a/builder/azure/arm/builder.go
+++ b/builder/azure/arm/builder.go
@@ -225,7 +225,7 @@ func (b *Builder) Run(ctx context.Context, ui packersdk.Ui, hook packersdk.Hook)
 			},
 			NewStepCreateResourceGroup(azureClient, ui),
 			NewStepValidateTemplate(azureClient, ui, &b.config, deploymentName, GetVirtualMachineDeployment),
-			NewStepDeployTemplate(azureClient, ui, &b.config, deploymentName, GetVirtualMachineDeployment, true),
+			NewStepDeployTemplate(azureClient, ui, &b.config, deploymentName, GetVirtualMachineDeployment, VirtualMachineTemplate),
 			NewStepGetIPAddress(azureClient, ui, endpointConnectType),
 			&communicator.StepConnectSSH{
 				Config:    &b.config.Comm,
@@ -257,7 +257,7 @@ func (b *Builder) Run(ctx context.Context, ui packersdk.Ui, hook packersdk.Hook)
 			keyVaultDeploymentName := b.stateBag.Get(constants.ArmKeyVaultDeploymentName).(string)
 			steps = append(steps,
 				NewStepValidateTemplate(azureClient, ui, &b.config, keyVaultDeploymentName, GetCommunicatorSpecificKeyVaultDeployment),
-				NewStepDeployTemplate(azureClient, ui, &b.config, keyVaultDeploymentName, GetCommunicatorSpecificKeyVaultDeployment, false),
+				NewStepDeployTemplate(azureClient, ui, &b.config, keyVaultDeploymentName, GetCommunicatorSpecificKeyVaultDeployment, KeyVaultTemplate),
 			)
 		} else if b.config.Comm.Type == "winrm" {
 			steps = append(steps, NewStepCertificateInKeyVault(&azureClient.VaultClient, ui, &b.config, b.config.winrmCertificate))
@@ -281,7 +281,7 @@ func (b *Builder) Run(ctx context.Context, ui packersdk.Ui, hook packersdk.Hook)
 			NewStepGetCertificate(azureClient, ui),
 			NewStepSetCertificate(&b.config, ui),
 			NewStepValidateTemplate(azureClient, ui, &b.config, deploymentName, GetVirtualMachineDeployment),
-			NewStepDeployTemplate(azureClient, ui, &b.config, deploymentName, GetVirtualMachineDeployment, true),
+			NewStepDeployTemplate(azureClient, ui, &b.config, deploymentName, GetVirtualMachineDeployment, VirtualMachineTemplate),
 			NewStepGetIPAddress(azureClient, ui, endpointConnectType),
 		)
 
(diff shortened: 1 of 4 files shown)


builder/azure/arm/step_deploy_template.go
```diff
@@ -299,9 +309,9 @@ func (s *StepDeployTemplate) deleteDeploymentResources(ctx context.Context, depl
 					resourceName,
 					resourceGroupName)
 				if err != nil {
-					s.say(fmt.Sprintf("Error deleting resource. Will retry.\n"+
```
JenGoldstrich, line 302: We regularly fail to delete attached network resources related to the resource being still attached, this leads to a lot of spam in the output, here I've changed this to only log the error when we fail for real, since almost every build I run fails this check at least once for at least two different resources.

builder/azure/arm/builder.go
```diff
@@ -281,7 +281,7 @@ func (b *Builder) Run(ctx context.Context, ui packersdk.Ui, hook packersdk.Hook)
 			NewStepGetCertificate(azureClient, ui),
 			NewStepSetCertificate(&b.config, ui),
 			NewStepValidateTemplate(azureClient, ui, &b.config, deploymentName, GetVirtualMachineDeployment),
-			NewStepDeployTemplate(azureClient, ui, &b.config, deploymentName, GetVirtualMachineDeployment),
+			NewStepDeployTemplate(azureClient, ui, &b.config, deploymentName, GetVirtualMachineDeployment, true),
```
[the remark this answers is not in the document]
nywilken: Another option, which I think helps readability, is to add an enum the to the step_deploy_template that can act as the last parameter, which will also help make the StepDeployTemplate step easier to grok, without having to look into its implementation.


```go
type DeploymentTemplateType int
const(
  VirtualMachineTemplate DeploymentTemplateType = iota
  KeyVaultTemplate
)
```

Then you can change the function to take in that type

```go
func NewStepDeployTemplate(client *AzureClient, ui packersdk.Ui, config *Config, deploymentName string, factory templateFactoryFunc, templateType DeploymentTemplateType ) *StepDeployTemplate {
	var step = &StepDeployTemplate{
		client:                     client,
		say:                        func(message string) { ui.Say(message) },
		error:                      func(e error) { ui.Error(e.Error()) },
		config:                     config,
		factory:                    factory,
		name:                       deploymentName,
		deploymentType:             templateType,
	}
```

```go
func (s *StepDeployTemplate) Cleanup(state multistep.StateBag){

//redacted
if s.deploymentTyp == KeyVaultTemplate {
	ui.Say("\nDeleting KeyVault created during build")
		err := s.delete(context.TODO(), deploymentName, resourceGroupName)
		if err != nil {
			s.say("Failed to delete the managed Additional Disk!")
			s.reportIfError(err, resourceGroupName)
		}
        return
}
// otherwise handle the VirtualMachine deletion case.

File: builder/azure/arm/step_deploy_template.go [hunk 3]
```diff
@@ -71,53 +73,61 @@ func (s *StepDeployTemplate) Cleanup(state multistep.StateBag) {
...
-	if imageType == "" && imageName == "" {
-		ui.Error(fmt.Sprintf("Failed to find temporary OS disk on VM.  Please delete manually.\n\n"+
-			"VM Name: %s\n"+
-			"Error: %s", computeName, err))
-		return
-	}
-	if !state.Get(constants.ArmKeepOSDisk).(bool) {
-		ui.Say(fmt.Sprintf(" Deleting -> %s : '%s'", imageType, imageName))
-		err = s.deleteDisk(context.TODO(), imageName, resourceGroupName, isManagedDisk)
+	if s.isVirtualMachineDeployment {
```
nywilken, line 78: Since the keyvault case is the guard case I would suggest checking if isVirtualMachineDeployment is false, run the key vault deletion and return earlier. Making the vm deletion the default logic.

builder/azure/arm/step_deploy_template_test.go
```diff
@@ -114,7 +117,7 @@ func TestStepDeployTemplateDeleteImageShouldFailWithInvalidImage(t *testing.T) {
 
 func TestStepDeployTemplateCleanupShouldDeleteManagedOSImageInExistingResourceGroup(t *testing.T) {
 	var deleteDiskCounter = 0
-	var testSubject = createTestStepDeployTemplateDeleteOSImage(&deleteDiskCounter)
+	var testSubject = createTestStepDeployTemplateDeleteOSImage(&deleteDiskCounter, true)
```
nywilken, line 120: Same thing here it is not clear what the boolean is for. But here you can create a variable initialized to true.

builder/azure/arm/step_deploy_template_test.go
```diff
@@ -184,6 +187,24 @@ func TestStepDeployTemplateCleanupShouldVHDOSImageInTemporaryResourceGroup(t *te
 	}
 }
 
+func TestStepDeployTemplateCleanupShouldNotDeleteDiskForKeyVaultDeployments(t *testing.T) {
```
nywilken, line 190: Nice test

nywilken (commented):
Hi there overall this looks good. I made a few comments on the boolean not being immediately clear. Take a look at the alternative suggestions and let me know what you think. 

I do suggest inverting the condition in the cleanup step to key if it is a keyVault deployment and return early as opposed to using the if/else statement.

builder/azure/arm/builder.go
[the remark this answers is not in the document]
JenGoldstrich: Implemented this suggestion, thanks for the feedback Wilken!

nywilken (approved):
Looks great! 🚢

PR #297, opened 2023-05-01
Fix Shared Image Gallery Sourced builds HCP Packer Ancestry and  add SIG Acceptance Tests

JenGoldstrich:
Before this PR the Source Image ID for a SIG sourced build would be the source VM that the Shared Image Gallery image version was built from, rather than the Shared Image Gallery image version itself.

```
SourceImageName: '/subscriptions/1f90521a-24f6-4758-ac3d-88d869fb0bf5/resourceGroups/pkr-Resource-Group-p6ehd2pbol/providers/Microsoft.Compute/virtualMachines/pkrvmp6ehd2pbol
```

After this PR the source image ID for a SIG sourced build is the parent SIG image version
```
-> SourceImageName: '/subscriptions/{subscription}/resourceGroups/jenna/providers/Microsoft.Compute/galleries/hcp_packer_ancestry/images/parent-image/versions/4.0.0'
```

For builds that are directly published to SIG, the plugin sends the SIG ID to HCP Packer as the Image ID which makes sense and works end to end with Terraform currently, thus we should use this ID as our `SourceImageName` the field which Packer uses to determine parent images

This PR also introduces SIG acceptance tests to the ARM builder acceptance tests.

### DIFF
--- a/builder/azure/arm/builder_acc_test.go
+++ b/builder/azure/arm/builder_acc_test.go
@@ -45,6 +45,10 @@ import (
 const DeviceLoginAcceptanceTest = "DEVICELOGIN_TEST"
 
 func TestBuilderAcc_ARM64SpecializedLinuxSIG(t *testing.T) {
+	if os.Getenv("AZURE_CLI_AUTH") == "" {
+		t.Skip("Azure CLI Acceptance tests skipped unless env 'AZURE_CLI_AUTH' is set, and an active `az login` session has been established")
+		return
+	}
 	acctest.TestPlugin(t, &acctest.PluginTestCase{
 		Name:     "test-specialized-linux-sig",
 		Type:     "azure-arm",
@@ -78,6 +82,10 @@ func TestBuilderAcc_ARM64SpecializedLinuxSIG(t *testing.T) {
 }
 
 func TestBuilderAcc_WindowsSIG(t *testing.T) {
+	if os.Getenv("AZURE_CLI_AUTH") == "" {
+		t.Skip("Azure CLI Acceptance tests skipped unless env 'AZURE_CLI_AUTH' is set, and an active `az login` session has been established")
+		return
+	}
 	acctest.TestPlugin(t, &acctest.PluginTestCase{
 		Name:     "test-windows-sig",
 		Type:     "azure-arm",
@@ -353,7 +361,7 @@ func createTestAzureClient(t *testing.T) AzureClient {
 		spnCloud,
 		spnKeyVault)
 	if err != nil {
-		t.Fatalf("failed to create azure client: %s", err)
+		t.Fatalf("failed to create test azure client: %s", err)
 	}
 	return *azureClient
 }
--- a/builder/azure/arm/builder_acc_test.go
+++ b/builder/azure/arm/builder_acc_test.go
@@ -46,6 +46,39 @@ import (
 
 const DeviceLoginAcceptanceTest = "DEVICELOGIN_TEST"
 
+func TestBuilderAcc_WindowsSIG(t *testing.T) {
+	b := Builder{}
+	_, _, _ = b.Prepare()
+	acctest.TestPlugin(t, &acctest.PluginTestCase{
+		Name:     "test-windows-sig",
+		Type:     "azure-arm",
+		Template: testBuilderAccSIGDiskWindows,
+		Setup: func() error {
+			createSharedImageGalleryDefinition(t, &b, CreateSharedImageGalleryDefinitionParameters{
+				galleryImageName: "windows-sig",
+				imageSku:         "2012-R2-Datacenter",
+				imageOffer:       "WindowsServer",
+				imagePublisher:   "MicrosoftWindowsServer",
+				isX64:            true,
+				isWindows:        true,
+			})
+			return nil
+		},
+		Check: func(buildCommand *exec.Cmd, logfile string) error {
+			if buildCommand.ProcessState != nil {
+				if buildCommand.ProcessState.ExitCode() != 0 {
+					return fmt.Errorf("Bad exit code. Logfile: %s", logfile)
+				}
+			}
+			return nil
+		},
+		Teardown: func() error {
+			deleteSharedImageGalleryDefinition(t, &b, "windows-sig")
+			return nil
+		},
+	})
+}
+
 func TestBuilderAcc_ManagedDisk_Windows(t *testing.T) {
 	acctest.TestPlugin(t, &acctest.PluginTestCase{
 		Name:     "test-azure-managedisk-windows",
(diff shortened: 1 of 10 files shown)


builder/azure/arm/builder_acc_test.go
```diff
@@ -293,63 +315,76 @@ func checkTemporaryGroupDeleted(t *testing.T, b *Builder) {
...
-	// Validate resource group has been deleted
-	_, err = azureClient.GroupsClient.Get(context.Background(), b.config.tmpResourceGroupName)
-	if err == nil || !resourceNotFound(err) {
-		t.Fatalf("failed validating resource group deletion: %s", err)
+	err = versionFuture.WaitForCompletionRef(context.TODO(), azureClient.GalleryImageVersionsClient.Client)
+	if err != nil {
+		t.Fatalf("failed to delete Gallery %s: %s", galleryImageName, err)
+	}
+	// WaitForCompletionRef is unreliable
+	time.Sleep(5000)
```
lbajolet-hashicorp, line 376: This will likely work but is indeed a bit flaky, I presume we could periodically check for the status of the SIG with like a GET or some other method, and give it a more generous timeout like a minute? Still flaky if the platform is slower than usual for some reason, but likely a bit less than a 5s wait.

builder/azure/arm/builder_acc_test.go
```diff
@@ -432,6 +467,43 @@ const testBuilderAccManagedDiskWindows = `
 }
 `
 
+const testBuilderAccSIGDiskWindows = `
+{
```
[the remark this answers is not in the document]
JenGoldstrich: Yeah absolutely, good point, I'll change that

builder/azure/arm/step_get_source_image_name.go
```diff
@@ -36,29 +34,9 @@ func (s *StepGetSourceImageName) Run(ctx context.Context, state multistep.StateB
...
-		image, err := client.Get(ctx, s.config.SharedGallery.ResourceGroup,
-			s.config.SharedGallery.GalleryName, s.config.SharedGallery.ImageName, s.config.SharedGallery.ImageVersion, "")
-
-		if err != nil {
-			log.Println("[TRACE] unable to derive managed image URL for shared gallery version image")
-			s.GeneratedData.Put("SourceImageName", "ERR_SOURCE_IMAGE_NAME_NOT_FOUND")
-			return multistep.ActionContinue
-		}
-
-		if image.GalleryImageVersionProperties != nil && image.GalleryImageVersionProperties.StorageProfile != nil &&
```
JenGoldstrich, line 51: When we wrote this logic you couldn't directly publish a shared image gallery image version from Packer, now you can, when you set that HCP Packer doesn't expect the Source ID of the parent Gallery Image Version (this would be like, the specific Windows marketplate image), for the parent, it expects the gallery image itself, because of that we can just infer the SIG resource ID and don't need to make the extra request

builder/azure/arm/builder_acc_test.go
JenGoldstrich: I resolved this by using a retry from the packer-plugin-sdk

lbajolet-hashicorp (approved):
LGTM!

builder/azure/arm/builder_acc_test.go
```diff
@@ -263,26 +320,28 @@ func TestBuilderAcc_rsaSHA2OnlyServer(t *testing.T) {
...
 
-//nolint
-func checkTemporaryGroupDeleted(t *testing.T, b *Builder) {
+func createTestAzureClient(t *testing.T) AzureClient {
+	b := Builder{}
+	_, _, _ = b.Prepare()
 	ui := testUi()
-
+	// Use CLI auth for our test client
+	b.config.ClientConfig.UseAzureCLIAuth = true
```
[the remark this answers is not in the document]
JenGoldstrich: Hey Wilken, good catch on 1 I need to add 
something like
```	
if os.Getenv("AZURE_CLI_AUTH") == "" {
		t.Skip("Azure CLI Acceptance tests skipped unless env 'AZURE_CLI_AUTH' is set, and an active `az login` session has been established")
		return
	}
```

2.) We have other tests that test env variables, they are all defined in their respective packer templates, I just wanted to pick one type of auth for this setup/tear down rather than orchestrating something more complicated that checked for different types of provided auth, and CLI seemed the simplest for now

builder/azure/arm/step_get_source_image_name.go
```diff
@@ -51,9 +63,16 @@ func (s *StepGetSourceImageName) Run(ctx context.Context, state multistep.StateB
...
+			sourceID := *image.GalleryImageVersionProperties.StorageProfile.Source.ID
+			isSIGSourcedFromManagedImage, _ := regexp.MatchString("/subscriptions/[^/]*/resourceGroups/[^/]*/providers/Microsoft.Compute/images/[^/]*$", sourceID)
+			// If the Source SIG Image Version does not have its StorageProfile.Source set to a Managed Image, that means the image was directly sourced from a VM
+			// Use the SIG ID itself if there isn't a managed image that the SIG was created from
+			if !isSIGSourcedFromManagedImage {
+				sourceID = *image.ID
+			}
+
+			s.say(fmt.Sprintf(" -> SourceImageName: '%s'", sourceID))
+			s.GeneratedData.Put("SourceImageName", sourceID)
```
nywilken, line 75: What is Image.ID here? How is that different from sourceID?

I think variables can be changed here to provide a little more context, as it was a little hard for me to understand. 
```go
			sigResourceID := *image.ID
			sourceImageResourceID := *image.GalleryImageVersionProperties.StorageProfile.Source.ID
			isSIGSourcedFromManagedImage, _ := regexp.MatchString("/subscriptions/[^/]*/resourceGroups/[^/]*/providers/Microsoft.Compute/images/[^/]*$", sourceID)
			// If the Source SIG Image Version does not have its StorageProfile.Source set to a Managed Image, that means the image was directly sourced from a VM
			// Use the SIG ID itself if there isn't a managed image that the SIG was created from
			if !isSIGSourcedFromManagedImage {
				sourceImageResourceID = sigResourceID
			}

			s.say(fmt.Sprintf(" -> SourceImageName: '%s'", sourceImageResourceID))
			s.GeneratedData.Put("SourceImageName", sourceImageResourceID)
```

builder/azure/arm/builder_acc_test.go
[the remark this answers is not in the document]
nywilken: Got it. I thought this was a requirement for all tests.

nywilken (commented):
Overall this looks good. I will run the acceptance tests and let you know if I run into any issues.

builder/azure/arm/builder_acc_test.go
```diff
@@ -26,22 +29,93 @@ import (
...
-	"github.com/Azure/go-autorest/autorest/azure/auth"
+	"github.com/Azure/azure-sdk-for-go/services/compute/mgmt/2021-11-01/compute"
 	"github.com/hashicorp/packer-plugin-sdk/acctest"
 	packersdk "github.com/hashicorp/packer-plugin-sdk/packer"
+	"github.com/hashicorp/packer-plugin-sdk/retry"
 )
 
 const DeviceLoginAcceptanceTest = "DEVICELOGIN_TEST"
 
+func TestBuilderAcc_ARM64SpecializedLinuxSIG(t *testing.T) {
```
nywilken, line 47: For ease of testing I suggest renaming these test a to include SIG or SharedImageGallery
```suggestion
func TestBuilderAcc_SharedImageGallery_ARM64SpecializedLinuxSIG(t *testing.T) {
```

This way we can run `PACKER_ACC=1 go test ./... -run= 'TestBuilderAcc_SharedImageGallery'` or just  `PACKER_ACC=1 go test ./... -run= 'SharedImageGallery'`

nywilken (approved):
Overall all this looks good. I approved. I am running the test again with the updates.

Issue #250, opened 2022-11-16
Connect to Azure Storage account using Managed identity and not Account key.


SebastianClaesson:
#### Community Note

Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the [original issue](https://github.com/hashicorp/packer-plugin-azure/issues/250#issue-1451160212) to help the community and maintainers prioritize this request.
Please do not leave "+1" or "me too" comments, they generate extra noise for issue followers and do not help prioritize the request.
If you are interested in working on this issue or have submitted a pull request, please leave a comment.

#### Description

The Azure ARM VHD plugin for packer will try to use the account key instead of leveraging the Managed Identity of the VM/VMSS.
In the [azure_client.go](https://github.com/hashicorp/packer-plugin-azure/blob/main/builder/azure/arm/azure_client.go#L271) on row 279 is the definition of the Azure Storage Account connection using an account key.
There could be implemented a check to see if client id/secret is set or if another variable (use_azure_msi?) is set, to support true managed identity integration to interact with the storage account, thus excluding the need for Account keys/SAS tokens, and only allow Azure AD integration.

#### Use Case(s)

To allow "true" managed identity integration when trying to create a golden image and upload it as a VHD.
Microsoft recommendations are to disallow the use of Account keys.

#### Potential References

How to set the Managed identity using Azure SDK for go.
https://pkg.go.dev/github.com/Azure/azure-sdk-for-go/sdk/azidentity#DefaultAzureCredential

Example:

``` go 
func Example_client_NewClient() {
	// this example uses Azure Active Directory (AAD) to authenticate with Azure Blob Storage
	accountName, ok := os.LookupEnv("AZURE_STORAGE_ACCOUNT_NAME")
	if !ok {
		panic("AZURE_STORAGE_ACCOUNT_NAME could not be found")
	}
	serviceURL := fmt.Sprintf("https://%s.blob.core.windows.net/", accountName)

	// https://pkg.go.dev/github.com/Azure/azure-sdk-for-go/sdk/azidentity#DefaultAzureCredential
	cred, err := azidentity.NewDefaultAzureCredential(nil)
	handleError(err)

	client, err := azblob.NewClient(serviceURL, cred, nil)
	handleError(err)

	fmt.Println(client.URL())
} 
```

Issue #243, opened 2022-10-14
[Notice] V2 Azure Plugin - Changing SDKs


JenGoldstrich:
The version of the Azure SDK for Go the Packer Plugin relies on was deprecated in March https://azure.microsoft.com/en-us/updates/support-for-azure-sdk-libraries-that-do-not-conform-to-our-current-azure-sdk-guidelines-will-be-retired-as-of-31-march-2023/, to handle this deprecation we have decided to switch to the same SDK that the Terraform AzureRM Provider uses, github.com/hashicorp/go-azure-sdk.  This handles a lot of authentication logic for us and allows us to quickly add a lot of new highly requested features, like OIDC.   We plan to label this release as v2.0.0  .

The SDK change itself largely does not change what API endpoints we call during builds and as such should not require changes for most users, we currently have a beta release containing the changes only for the ARM Builder https://github.com/JenGoldstrich/packer-plugin-azure/releases/tag/v2.0.0-betarc1 which users may test if they are concerned about this breaking their workflows, this only contains the changes for the ARM builder, and not DevTestLabs and CHroot yet.  We are working on that next and will then release v2, the full list of current changes is available In the beta release notes, but it’s worth noting it is not the full list of changes in the v2 release.  

## Full List of current breaking changes  
1. We no longer connect to Storage Accounts via an account key but instead use a managed identity, as requested in hashicorp#250. As a consequence, the Teardown operation after a successful VHD build will fail unless the managed identity has permission to delete from the storage account. VHD builds will still succeed without this access, but they will leave an orphaned OS blob in their storage account.
2. Read Only SAS Uris that generated for VHD builds and managed image builds when sttting keep_os_disk to true, are no longer generated by default to mitigate security concerns. The Microsoft team does not recommend generating these SAS URIs as well.
3. Device Code authentication has been removed in favor of the recommend approach of using az login --use-device-code instead to replicate this functionality.

PR #306, opened 2023-05-31
SDK Migration - V2 Beta

JenGoldstrich:
### Description

This PR switches from the ARM Builder from the ["track 1" SDK we've been using](github.com/Azure/azure-sdk-for-go) to an SDK maintained by the [Terraform Azure Provider team](github.com/hashicorp/Terraform-Provider-AzureRM), this new SDK is located [here](github.com/hashicorp/go-azure-sdk).  It has drop in support for track 2 SDKs to help us address the deprecation of the existing SDK with fewer breaking changes.  As the new SDK adds track 2 support we can migrate them over time

Related to https://github.com/hashicorp/packer-plugin-azure/issues/243

### Breaking Changes
- All ARM builder functionality moved to github.com/hashicorp/go-azure-sdk
- Address https://github.com/hashicorp/packer-plugin-azure/issues/250 for the ARM Builder as we are currently considering removing the Read Only SAS's that are generated during VHD builds.  And no longer create a connection to the storage account via an account key
- Device Code authentication is removed, we are reccomending using `az login --use-device-code` instead to replicate this functionality

### Other Changes
- Removed 5 minute expiration check on JWT tokens, as OIDC tokens don't work with this expiration check

### What's not done
- Usage of PollingDurationTimeout as the actual polling duration timeout consistently throughout new code
- DTL Builder and CHroot are not tested or guaranteed to work with this beta release, they will be migrated after we have validated the migration of the ARM builder and gotten feedback from users
- The new SDK supports has some endpoints on track 2, these endpoints require the creation of new request/response handlers that will have to be added before the full v2 release


builder/azure/common/client/config.go
```diff
@@ -224,9 +224,9 @@ func (c Config) Validate(errs *packersdk.MultiError) {
 		if err != nil {
 			errs = packersdk.MultiErrorAppend(errs, fmt.Errorf("client_jwt is not a JWT: %v", err))
 		} else {
-			if claims.ExpiresAt < time.Now().Add(5*time.Minute).Unix() {
-				errs = packersdk.MultiErrorAppend(errs, fmt.Errorf("client_jwt will expire within 5 minutes, please use a JWT that is valid for at least 5 minutes"))
-			}
+			//if claims.ExpiresAt < time.Now().Add(5*time.Minute).Unix() {
```
JenGoldstrich, line 227: OIDC Access Tokens have a shorter validatity, so throwing this error prevents proper OIDC support, removing it fixes this, I have a slack link with some more context from the provider folks if needed

builder/azure/common/client/config.go
```diff
@@ -109,6 +109,11 @@ func (c *Config) CloudEnvironment() *azure.Environment {
 	return c.cloudEnvironment
 }
 
+func (c *Config) AuthType() string {
+	return c.authType
+}
+
+// TODO I still need to port this to the new SDK
```
JenGoldstrich, line 116: This logic is used by every builder, I'll port is towards the end of the process and leave this todo until then

builder/azure/arm/step_deploy_template.go
```diff
@@ -305,12 +314,13 @@ func (s *StepDeployTemplate) deleteDeploymentResources(ctx context.Context, depl
 			defer wg.Done()
 			retryConfig := retry.Config{
 				Tries:      10,
-				RetryDelay: (&retry.Backoff{InitialBackoff: 10 * time.Second, MaxBackoff: 600 * time.Second, Multiplier: 2}).Linear,
+				RetryDelay: (&retry.Backoff{InitialBackoff: 5 * time.Second, MaxBackoff: 60 * time.Second, Multiplier: 1.5}).Linear,
```
JenGoldstrich, line 317: This change seems to speed up builds quite a bit, if the delete failed a few times here we could end up waiting for quite a long time, this cuts off the max backoff at 1 minute, and lowers the multiplier that increases the backoff, anecdotally acceptance tests feel way faster, and the OIDC test build only took 5 minutes.

builder/azure/arm/step_deploy_template.go
```diff
@@ -107,11 +107,11 @@ func (s *StepDeployTemplate) Cleanup(state multistep.StateBag) {
 		// delete the disk as the image request will return a 404
 		computeName := state.Get(constants.ArmComputeName).(string)
 		isManagedDisk := state.Get(constants.ArmIsManagedImage).(bool)
-		imageType, imageName, err := s.disk(context.TODO(), subscriptionId, resourceGroupName, computeName)
+		imageType, imageName, err := s.disk(ctx, subscriptionId, resourceGroupName, computeName)
```
nywilken, line 110: Nice catch

builder/azure/arm/builder.go
```diff
@@ -65,7 +65,7 @@ func (b *Builder) Run(ctx context.Context, ui packersdk.Ui, hook packersdk.Hook)
 
 	ui.Say("Running builder ...")
 
-	ctx, cancel := context.WithCancel(ctx)
+	ctx, cancel := context.WithTimeout(ctx, time.Minute*60)
```
nywilken, line 68: Does it make sense to make this timeout configurable via some configuration option?

builder/azure/arm/builder.go
```diff
@@ -111,31 +116,31 @@ func (b *Builder) Run(ctx context.Context, ui packersdk.Ui, hook packersdk.Hook)
 	if err := resolver.Resolve(&b.config); err != nil {
 		return nil, err
 	}
+
 	if b.config.ClientConfig.ObjectID == "" {
-		b.config.ClientConfig.ObjectID = getObjectIdFromToken(ui, spnCloud)
+		b.config.ClientConfig.ObjectID = *objectID
```
nywilken, line 121: Is there ever a case where objectID is nil and this throws a nil pointer deference error?

builder/azure/arm/builder.go
```diff
@@ -111,31 +116,31 @@ func (b *Builder) Run(ctx context.Context, ui packersdk.Ui, hook packersdk.Hook)
...
+		imageId := hashiImagesSDK.NewImageID(b.config.ClientConfig.SubscriptionID, b.config.ManagedImageResourceGroupName, b.config.ManagedImageName)
+		_, err = azureClient.ImagesClient.Get(ctx, imageId, hashiImagesSDK.DefaultGetOperationOptions())
 		if err == nil {
 			if b.config.PackerForce {
 				ui.Say(fmt.Sprintf("the managed image named %s already exists, but deleting it due to -force flag", b.config.ManagedImageName))
-				f, err := azureClient.ImagesClient.Delete(ctx, b.config.ManagedImageResourceGroupName, b.config.ManagedImageName)
-				if err == nil {
-					err = f.WaitForCompletionRef(ctx, azureClient.ImagesClient.Client)
-				}
+				err := azureClient.ImagesClient.DeleteThenPoll(ctx, imageId)
```
nywilken, line 143: Nice

builder/azure/arm/resource_resolver.go
```diff
@@ -75,41 +76,34 @@ func getResourceGroupNameFromId(id string) string {
 	return xs[4]
 }
 
-func findManagedImageByName(client *AzureClient, name, resourceGroupName string) (*compute.Image, error) {
-	images, err := client.ImagesClient.ListByResourceGroupComplete(context.TODO(), resourceGroupName)
+func findManagedImageByName(client *AzureClient, name, subscriptionId, resourceGroupName string) (*hashiImagesSDK.Image, error) {
+	id := commonids.NewResourceGroupID(subscriptionId, resourceGroupName)
+	images, err := client.ImagesClient.ListByResourceGroupComplete(context.TODO(), id)
```
nywilken, line 81: Not really part of the review but I'm curious why the code has context.TODO() as opposed to reusing the ctx created during runtime? Guessing there is no context being passed in would be my first guess.

go.mod
```diff
@@ -96,23 +109,30 @@ require (
...
 	google.golang.org/protobuf v1.28.1 // indirect
+	gopkg.in/check.v1 v1.0.0-20201130134442-10cb98267c6c // indirect
 	gopkg.in/square/go-jose.v2 v2.6.0 // indirect
 	gopkg.in/yaml.v2 v2.3.0 // indirect
-	gopkg.in/yaml.v3 v3.0.0-20210107192922-496545a6307b // indirect
+	gopkg.in/yaml.v3 v3.0.1 // indirect
+	software.sslmate.com/src/go-pkcs12 v0.2.0 // indirect
 )
 
+replace github.com/zclconf/go-cty => github.com/nywilken/go-cty v0.0.0-20230418194721-b97b674b17ec
```
nywilken, line 135: A new temporary fork has been generated 
```go
replace github.com/zclconf/go-cty => github.com/nywilken/go-cty v0.0.0-20230418194721-b97b674b17ec
```

builder/azure/arm/resource_resolver.go
JenGoldstrich: Yes! So the context for the clients is mostly used for canceling and polling deadlines, I have not passed it through to here yet since it was only one get request used outside of a step, happy to add a real context here if you think it'd be better

go.mod
JenGoldstrich: Added in newest commit

builder/azure/arm/step_capture_image.go
```diff
@@ -44,89 +45,90 @@ func NewStepCaptureImage(client *AzureClient, ui packersdk.Ui) *StepCaptureImage
...
-			err = s.captureVhd(ctx, resourceGroupName, computeName, vmCaptureParameters)
+			return multistep.ActionHalt
 		}
-	}
-	if err != nil {
-		state.Put(constants.Error, err)
-		s.error(err)
 
-		return multistep.ActionHalt
+		// HACK(chrboum): I do not like this.  The capture method should be returning this value
```
lbajolet-hashicorp, line 121: Who's `cherboum` and `paulmey`? Are they working on the SDK?
Oh nevermind that's an older comment that was moved, still, is this hack still relevant with the refactor?

builder/azure/arm/step_create_resource_group.go
```diff
@@ -46,13 +48,17 @@ func (s *StepCreateResourceGroup) createResourceGroup(ctx context.Context, resou
...
+	exists, err := s.client.ResourceGroupsClient.Get(ctx, id)
 	if err != nil {
+		if exists.HttpResponse.StatusCode == 404 {
+			return false, nil
+		}
 		s.say(s.client.LastError.Error())
 	}
 
-	return exists.Response.StatusCode != 404, err
+	return exists.HttpResponse.StatusCode != 404, nil
```
lbajolet-hashicorp, line 62: Does that happen? If the error might be linked to a `404`, should we do the check in both places?

builder/azure/arm/step_create_resource_group.go
```diff
@@ -46,13 +48,17 @@ func (s *StepCreateResourceGroup) createResourceGroup(ctx context.Context, resou
 	return err
 }
 
-func (s *StepCreateResourceGroup) doesResourceGroupExist(ctx context.Context, resourceGroupName string) (bool, error) {
-	exists, err := s.client.GroupsClient.CheckExistence(ctx, resourceGroupName)
+func (s *StepCreateResourceGroup) doesResourceGroupExist(ctx context.Context, subscriptionId string, resourceGroupName string) (bool, error) {
+	id := commonids.NewResourceGroupID(subscriptionId, resourceGroupName)
+	exists, err := s.client.ResourceGroupsClient.Get(ctx, id)
 	if err != nil {
+		if exists.HttpResponse.StatusCode == 404 {
```
lbajolet-hashicorp, line 56: Is there a chance that `HttpResponse` might be `nil` here? If so this will panic.

lbajolet-hashicorp (commented):
Reviewed ~half for now, will continue this tomorrow, overall there's a ton of changes and I left more questions than proper comments, but this looks good to me at first glance.
I notice that only the `arm` builder is updated now, will you refactor the rest later?

builder/azure/arm/azure_client.go
```diff
@@ -131,167 +137,144 @@ func byConcatDecorators(decorators ...autorest.RespondDecorator) autorest.Respon
...
+	}
 
-	azureClient.DisksClient = compute.NewDisksClientWithBaseURI(cloud.ResourceManagerEndpoint, subscriptionID)
-	azureClient.DisksClient.Authorizer = autorest.NewBearerAuthorizer(servicePrincipalToken)
-	azureClient.DisksClient.RequestInspector = withInspection(maxlen)
-	azureClient.DisksClient.ResponseInspector = byConcatDecorators(byInspecting(maxlen), errorCapture(azureClient))
-	azureClient.DisksClient.UserAgent = fmt.Sprintf("%s %s", useragent.String(version.AzurePluginVersion.FormattedVersion()), azureClient.DisksClient.UserAgent)
+	// Clients that have been ported to hashicorp/go-azure-sdk
+	azureClient.DisksClient = hashiDisksSDK.NewDisksClientWithBaseURI(cloud.ResourceManagerEndpoint)
+	azureClient.DisksClient.Client.Authorizer = authWrapper.AutorestAuthorizer(resourceManagerAuthorizer)
```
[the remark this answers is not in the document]
JenGoldstrich: So autorest is an intermediary library used in track 1 SDKs, hashicorp/go-azure-sdk is still migrating from auto-rest to its own transit layer.

This code has been converted to use a different auth layer than Autorest, but track 1 SDKs still rely on autorest, so we can use the helper method in autoWrapper to convert this for SDKs not converted, then each time a client is moved to track 2 we will just replace authWrapper.AutorestAuthorizer(resourceManagerAuthorizer) with resourceManagerAuthorizer, as that helper function just turns the "new" auth object into the old one

builder/azure/arm/step_capture_image.go
JenGoldstrich: Yes but I haven't fully vetted that, there are two helper methods we use in azure client that still rely on autorest specific libraries, these are working on the new SDK currently but I need to migrate them still

builder/azure/arm/artifact.go
```diff
@@ -56,7 +53,7 @@ type Artifact struct {
 	StateData map[string]interface{}
 }
 
-func NewManagedImageArtifact(osType, resourceGroup, name, location, id, osDiskSnapshotName, dataDiskSnapshotPrefix string, generatedData map[string]interface{}, keepOSDisk bool, template *CaptureTemplate, getSasUrl func(name string) string) (*Artifact, error) {
```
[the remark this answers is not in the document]
JenGoldstrich: Good question, a SAS URL is a link to an Azure Storage account resource with an embedded token, we create SAS tokens in VHD builds and when using managed image builds without deleting the disk I believe, however we've been asked by a lot of our users to remove this here https://github.com/hashicorp/packer-plugin-azure/issues/250, because of the security concerns of using these they are not recommended by Microsoft, we are doing them in every build.  To top it off the new SDK does not provide the same helper we currently use to generate these tokens.

So for v2 after chatting with Wilken, and given the above issue, I am proposing removing them, they can still be created by users outside of Packer, and this was never used in the resulting Azure Artifact, but was simply printed in the terminal output of the build.

Yes we should remove it from the DTL builder but I would like to hold of any changes to the DTL builder until we have signed off on the way we've handling this for the ARM builder

builder/azure/arm/resource_resolver.go
```diff
@@ -75,41 +76,34 @@ func getResourceGroupNameFromId(id string) string {
...
-			return nil, err
-		}
 	}
 
 	return nil, fmt.Errorf("Cannot find an image named '%s' in the resource group '%s'", name, resourceGroupName)
 }
 
-func findVirtualNetworkResourceGroup(client *AzureClient, name string) (string, error) {
-	virtualNetworks, err := client.VirtualNetworksClient.ListAllComplete(context.TODO())
+func findVirtualNetworkResourceGroup(client *AzureClient, subscriptionId, name string) (string, error) {
```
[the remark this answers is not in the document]
JenGoldstrich: Azure Subscriptions are user created groups where resources are organized. Every resource (VM, Managed Image) has a Subscription, an a Resource Group which are  sub-divisions of subscriptions.

In the old SDK we passed in the subscription ID into the Azure Client, in the new SDK they ask for the subscription ID on every call but the resulting API call is the same regardless.

builder/azure/arm/step_publish_to_shared_image_gallery.go
```diff
@@ -7,16 +7,16 @@ import (
...
+	hashiGalleryImageVersionsSDK "github.com/hashicorp/go-azure-sdk/resource-manager/compute/2022-03-03/galleryimageversions"
 	"github.com/hashicorp/packer-plugin-azure/builder/azure/common/constants"
 	"github.com/hashicorp/packer-plugin-sdk/multistep"
 	packersdk "github.com/hashicorp/packer-plugin-sdk/packer"
 )
 
 type StepPublishToSharedImageGallery struct {
 	client  *AzureClient
-	publish func(ctx context.Context, sourceID string, sharedImageGallery SharedImageGalleryDestination, miSGImageVersionEndOfLifeDate string, miSGImageVersionExcludeFromLatest bool, miSigReplicaCount int32, location string, diskEncryptionSetId string, tags map[string]*string) (string, error)
+	publish func(ctx context.Context, subscriptionID string, sourceID string, sharedImageGallery SharedImageGalleryDestination, miSGImageVersionEndOfLifeDate string, miSGImageVersionExcludeFromLatest bool, miSigReplicaCount int64, location string, diskEncryptionSetId string, tags map[string]string) (string, error)
```
lbajolet-hashicorp, line 19: Out of curiosity, what's the `mi` in `miSGImageVersion...` referring to?
Machine image maybe?

builder/azure/arm/template_factory_test.TestVirtualMachineDeployment13.approved.json
```diff
@@ -130,7 +130,7 @@
               "listeners": [
                 {
                   "certificateUrl": "",
-                  "protocol": "https"
+                  "protocol": "Https"
```
lbajolet-hashicorp, line 133: Weird casing here, is this voluntary?

builder/azure/common/pointer.go
```diff
@@ -0,0 +1,36 @@
+// Copyright (c) HashiCorp, Inc.
+// SPDX-License-Identifier: MPL-2.0
+
+//go:generate packer-sdc struct-markdown
+
+package common
+
+// StringPtr returns a pointer to the passed string.
+func StringPtr(s string) *string {
```
lbajolet-hashicorp, line 9: I'm surprised we have to make our own, isn't there something like this in the Azure SDK already?

lbajolet-hashicorp: Might be a good idea to push it to the Hashicorp Azure SDK if not, especially if there's a bunch of string pointers in the API calls.

builder/azure/common/template/template.go
```diff
@@ -4,8 +4,14 @@
 package template
```
lbajolet-hashicorp, line 4: Out of curiosity again, what is being templated here? I'm not sure I understand what's the use for this package.

builder/azure/common/pointer.go
JenGoldstrich: This was in the old SDK as an annoying helper tbh, and I didn't want to have to have a bunch of duplicated code.  The new SDK does not have this and tbh I wasn't sure how it'd be received as a PR so I felt this was simpler

builder/azure/arm/template_factory_test.TestVirtualMachineDeployment13.approved.json
JenGoldstrich: This is what the new SDK sends, they resolve the same but we do a file compare to check in the test that causes failures cause https != https, Azure interprets them the same

builder/azure/arm/step_publish_to_shared_image_gallery.go
JenGoldstrich: I guess so, not 100% sure tbh

builder/azure/arm/step_create_resource_group.go
JenGoldstrich: Yes but this was true in the old SDK and it never happened, definitely worth guarding against but unlikely to happen I believe

builder/azure/common/client/config.go
```diff
@@ -102,12 +104,73 @@ func (c *Config) SetDefaultValues() error {
...
+	if err != nil {
+		// fall back to old method of normalizing and looking up cloud names.
+		log.Printf(fmt.Sprintf("Error looking up environment using metadata host: %s. \n"+
+			"Falling back to hardcoded mechanism...", err.Error()))
+		lookup := map[string]string{
+			"CHINA":           "china",
+			"CHINACLOUD":      "china",
+			"AZURECHINACLOUD": "china",
+
+			// TODO Implement AzureGermanCloud
```
maxilampert, line 141: FYI Azure Germany doesn't exist anymore since October 2021. There are Azure regions in Germany but they are part of the general public cloud. Ref: [Azure Germany](https://learn.microsoft.com/en-us/previous-versions/azure/germany/germany-welcome)

PR #311, opened 2023-06-30
Add encryption at host flag to the arm template

kamilpi:
This is an additional parameter already existing in the arm template. We have restricted policies on our infrastructure where we are not able to create the virtual machine without Encryption at host enabled.

I know that it could be a dedicated subscription for the build process but this is the organization and certification requirement.

### DIFF
--- a/builder/azure/arm/config.go
+++ b/builder/azure/arm/config.go
@@ -562,6 +562,9 @@ type Config struct {
 	LicenseType string `mapstructure:"license_type" required:"false"`
 	// Specifies if Secure Boot and Trusted Launch is enabled for the Virtual Machine.
 	SecureBootEnabled bool `mapstructure:"secure_boot_enabled" required:"false"`
+	// Specifies if Encryption at host is enabled for the Virtual Machine.
+	// Requires enabling encryption at host in the Subscription read more [here](https://learn.microsoft.com/en-us/azure/virtual-machines/disks-enable-host-based-encryption-portal?tabs=azure-powershell)
+	EncryptionAtHost bool `mapstructure:"encryption_at_host" required:"false"`
 
 	// Specifies if vTPM (virtual Trusted Platform Module) and Trusted Launch is enabled for the Virtual Machine.
 	VTpmEnabled bool `mapstructure:"vtpm_enabled" required:"false"`
--- a/builder/azure/common/template/template_builder.go
+++ b/builder/azure/common/template/template_builder.go
@@ -501,7 +501,7 @@ func (s *TemplateBuilder) SetLicenseType(licenseType string) error {
 	return nil
 }
 
-func (s *TemplateBuilder) SetSecurityProfile(secureBootEnabled bool, vtpmEnabled bool) error {
+func (s *TemplateBuilder) SetSecurityProfile(secureBootEnabled bool, vtpmEnabled bool, encryptionAtHost bool) error {
 	s.setVariable("apiVersion", "2020-12-01") // Required for Trusted Launch
 	resource, err := s.getResourceByType(resourceVirtualMachine)
 	if err != nil {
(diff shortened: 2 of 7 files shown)


builder/azure/common/template/template_builder.go
```diff
@@ -513,6 +513,7 @@ func (s *TemplateBuilder) SetSecurityProfile(secureBootEnabled bool, vtpmEnabled
 	resource.Properties.SecurityProfile.SecurityType = compute.SecurityTypesTrustedLaunch
 	resource.Properties.SecurityProfile.UefiSettings.SecureBootEnabled = to.BoolPtr(secureBootEnabled)
 	resource.Properties.SecurityProfile.UefiSettings.VTpmEnabled = to.BoolPtr(vtpmEnabled)
+	resource.Properties.SecurityProfile.EncryptionAtHost = to.BoolPtr(encryptionAtHost)
```
JenGoldstrich, line 516: Can we add an if statement here to check if secureBootEnabled or vtpmEnabled is set, and then set the SecurityType/Uefi Settings, otherwise just set the EncryptionAtHost, like this

```
resource.Properties.SecurityProfile = &compute.SecurityProfile{}
if secureBootEnabled || vtpmEnabled {
	resource.Properties.SecurityProfile.UefiSettings = &compute.UefiSettings{}
	resource.Properties.SecurityProfile.SecurityType = compute.SecurityTypesTrustedLaunch
	resource.Properties.SecurityProfile.UefiSettings.SecureBootEnabled = to.BoolPtr(secureBootEnabled)
	resource.Properties.SecurityProfile.UefiSettings.VTpmEnabled = to.BoolPtr(vtpmEnabled)
	resource.Properties.SecurityProfile.EncryptionAtHost = to.BoolPtr(encryptionAtHost)
}
resource.Properties.SecurityProfile.EncryptionAtHost = to.BoolPtr(encryptionAtHost)

That way we aren't setting trusted launch unless the user enables it, I believe its valid to set EncryptionAtHost but not trusted launch.  When building a VHD with UbuntuServer 16.04-LTS as an example and this was failing due to being incompatible with Trusted Launch, but you can still set encrypt it at host

File: builder/azure/arm/config.go [hunk 2]
```diff
@@ -562,6 +562,8 @@ type Config struct {
 	LicenseType string `mapstructure:"license_type" required:"false"`
 	// Specifies if Secure Boot and Trusted Launch is enabled for the Virtual Machine.
 	SecureBootEnabled bool `mapstructure:"secure_boot_enabled" required:"false"`
+	// Specifies if Encryption at host is enabled for the Virtual Machine.
```
JenGoldstrich, line 565:
```suggestion
	// Specifies if Encryption at host is enabled for the Virtual Machine.
	// Requires enabling encryption at host in Azure read more [here](https://learn.microsoft.com/en-us/azure/virtual-machines/disks-enable-host-based-encryption-portal?tabs=azure-powershell)
```

JenGoldstrich (commented):
Hey @kamilpi thank you for this contribution, I added a comment about trusted launch but can you also take a look at the unit tests failing in the GitHub Checks? They are failing because your changes mean we generate a different Arm VM template than these unit tests expect, you can find these templates in `./builder/azure/arm/template_factory_test.TestVirtualMachineDeployment*.json` please let me know if you have any questions or need some help fixing these unit tests!

builder/azure/common/template/template_builder.go
kamilpi: Sure, I will do this but we do not need the double assignment:
```
	resource.Properties.SecurityProfile.EncryptionAtHost = to.BoolPtr(encryptionAtHost)
}
resource.Properties.SecurityProfile.EncryptionAtHost = to.BoolPtr(encryptionAtHost)

```

I will try to cover all tomorrow.

JenGoldstrich: Oops yes, great catch!

builder/azure/arm/config.go
JenGoldstrich: @kamilpi do you mind taking in this change as well? Afterwards you need to run `make generate` to generate the docs site from the config, let me know if you have any trouble with this as I can just commit the change to your fork

kamilpi: Done.

JenGoldstrich (approved):

JenGoldstrich:
Thanks for this contribution @kamilpi merging this now! We should have a release out containing this change in a few weeks

oliverlabs:
does this solve the Trusted Launch issue @JenGoldstrich ? :)

mala3000:
This breaks the build process when the feature is not enabled in the Azure subscription and the packer variable is not set, while it worked perfectly fine on 1.4.3.
If the variable is not set in the packer config, it shouldn't try to set it in Azure, the same as it was in 1.4.3.

Details=[{"code":"BadRequest","message":"{\r\n  \"error\": {\r\n    \"code\": \"InvalidParameter\",\r\n    \"message\": \"The property 'securityProfile.encryptionAtHost' is not valid because the 'Microsoft.Compute/EncryptionAtHost' feature is not enabled for this subscription.\",\r\n    \"target\": \"securityProfile.encryptionAtHost\"\r\n  }\r\n}"}]

PR #317, opened 2023-07-06
Parallelize the acceptance tests

JenGoldstrich:
None of these tests build conflicting resources, so we should force them to run in parallel.
Drastically increases speed of acceptance tests.


nywilken (approved):
LGTM

PR #327, opened 2023-08-07
[V2] Use PFX Certificates for Client Auth instead of PEM

JenGoldstrich:
In the new SDK we are using in v2 client certificates behave the way they do for the Terraform Provider https://registry.terraform.io/providers/hashicorp/azuread/latest/docs/guides/service_principal_client_certificate using .pfx files only. This also allows user to use password protected certificates which is preferred for security.  This will require users to generate new certificates sadly which is a breaking change

Adds `client_cert_password` field, as most pfx files are password protected and the SDK supports this

Removes `client_cert_timeout` field, which is not supported in the new SDK

This PR has a messy diff since its on top of my fork with all of V2 migration, but the main changers are in the azure_authorizer file, and in the common client config

Closes https://github.com/hashicorp/packer-plugin-azure/issues/46


lbajolet-hashicorp (commented):
Still gotta work through the V2 PR (up next!), but the cert logic looks good to me. The TODO test needs to be written still, I'll circle back on this when it's done

JenGoldstrich:
@lbajolet-hashicorp I messaged you in slack about this but just to close the loop that test is not planned to be written as part of this PR, and is intended to document a pre-exsisting missing test case we want to circle back on when we have acceptance tests orchestrated in CI with Terraform

nywilken (approved):
Change wise this looks good. Please rebase and drop the rest of v2 before merging. Or you can drop and retarget to the v2 branch so that things get merged in order. 

Since there are fields being removed and there is a new flow. Do we need to add more documentation for how to migrate to the new cert? 

To not create too much conflict with the docs. Please include upgrade notes in the release notes on what users should do to migrate to the new PFX certificate.

PR #328, opened 2023-08-08
Migration plugin docs to integration framework

nywilken:
Add scaffolding for [HashiCorp integrations](https://github.com/hashicorp/integrations)

### DIFF
--- a/GNUmakefile
+++ b/GNUmakefile
@@ -35,3 +35,8 @@ generate: install-packer-sdc
 	packer-sdc renderdocs -src ./docs -dst ./.docs -partials ./docs-partials
 	# checkout the .docs folder for a preview of the docs
 
+build-docs: install-packer-sdc
+	@if [ -d ".docs" ]; then rm -r ".docs"; fi
+	@packer-sdc renderdocs -src "docs" -partials docs-partials/ -dst ".docs/"
+	@./.web-docs/scripts/compile-to-webdocs.sh "." ".docs" ".web-docs" "hashicorp"
+	@rm -r ".docs"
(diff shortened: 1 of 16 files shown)


lbajolet-hashicorp (approved):
LGTM!

GNUmakefile
```diff
@@ -35,3 +35,8 @@ generate: install-packer-sdc
 	packer-sdc renderdocs -src ./docs -dst ./.docs -partials ./docs-partials
 	# checkout the .docs folder for a preview of the docs
 
+build-docs: install-packer-sdc
```
JenGoldstrich, line 38: Why do we still leave the older generate task in place, should we only use build-docs after this PR is merged?

BrandonRomano (approved):

GNUmakefile
nywilken: If you're referring to the release pipeline. We are keeping everything the same until we've cut over the plugins to production. Once live we can remove the doc.zip generation. But it should not affect anything if left in place.

docs/README.md
```diff
@@ -1,69 +1,89 @@
...
-Then, run [`packer init`](https://www.packer.io/docs/commands/init).
+To install this plugin, copy and paste this code into your Packer configuration, then run [`packer init`](https://www.packer.io/docs/commands/init).
 
 ```hcl
 packer {
   required_plugins {
     azure = {
-      version = ">= 1.4.2"
       source  = "github.com/hashicorp/azure"
+      version = "~> 1"
```
lbajolet-hashicorp, line 13: Now that v2 is out, we could update this block to point to this version I would think

nywilken: Updated.

PR #330, opened 2023-08-11
 [V2] Use PFX Certificates for Client Auth instead of PEM #327 

JenGoldstrich:
Duplicate of https://github.com/hashicorp/packer-plugin-azure/pull/327

In the new SDK we are using in v2 client certificates behave the way they do for the Terraform Provider https://registry.terraform.io/providers/hashicorp/azuread/latest/docs/guides/service_principal_client_certificate using .pfx files only. This also allows user to use password protected certificates which is preferred for security. This will require users to generate new certificates sadly which is a breaking change

Adds client_cert_password field, as most pfx files are password protected and the SDK supports this

Removes client_cert_timeout field, which is not supported in the new SDK

This PR has a messy diff since its on top of my fork with all of V2 migration, but the main changers are in the azure_authorizer file, and in the common client config

Closes https://github.com/hashicorp/packer-plugin-azure/issues/46


lbajolet-hashicorp (approved):
LGTM!

PR #335, opened 2023-08-21
Allow setting key vault secret name & Unify API versions

maxilampert:
This PR deals with two things.

1. Allow setting the key vault secret name by the 'build_key_vault_secret_name' parameter. The default value is still 'packerKeyVaultSecret' if the parameter is not specified.

2. Unify the ARM API versions to use there latest counterparts for compute and network resources globally and remove a lot of individual settings. Should not be breaking as the APIs are available in all clouds. Minor structural cleanup of the ARM templates.


JenGoldstrich:
Hey @maxilampert, generally this PR looks good but I'm curious what your use-case is for renaming the secret, the key vault secret is meant to be temporary and be deleted by the end of a build, since the key vault itself is also deleted.

maxilampert:
> Hey @maxilampert, generally this PR looks good but I'm curious what your use-case is for renaming the secret, the key vault secret is meant to be temporary and be deleted by the end of a build, since the key vault itself is also deleted.

Hey @JenGoldstrich, well in our case we use a fixed key vault that has been created beforehand and does not get deleted after the image build. We have some conventions in place for how secrets should be named. 
Also for a case of concurrent builds it could be useful to create separate secrets to not run into any issues.

JenGoldstrich (approved):
@maxilampert thanks for that context, although this makes me think that we should be deleting the key vault secret that we create when we don't delete the key vault, I believe that's an oversight more than intentional behavior.

I have created this issue to track that https://github.com/hashicorp/packer-plugin-azure/issues/343, if this would cause disruption to your workflow let me know, but I think it should be safe to do.  In the mean time I don't wanna continue to delay this PR though, and its a reasonable change to me.

I ran acceptance tests on this PR and tested a few manual builds while setting my secret name in a template, LGTM!

PR #337, opened 2023-08-28
Add `use_shallow_replication` field for Compute Gallery/SIG builds, fix replica count type

### LINKED ISSUE #313: FEATURE: Add Shallow option for publishing to ACG
Could you please add the `--replication-mode Shallow` support for publishing to ACG?

As per Az CLI documentation [here](https://learn.microsoft.com/en-us/azure/virtual-machines/shared-image-galleries?tabs=azure-cli#shallow-replication).

#### Community Note

#### Description

As per Az CLI documentation [here](https://learn.microsoft.com/en-us/azure/virtual-machines/shared-image-galleries?tabs=azure-cli#shallow-replication).

#### Use Case(s)

Faster iterations between Packer runs for developers! Average replication takes 15-20 minutes.

#### Potential configuration

```
```

### LINKED ISSUE #341: shared_image_gallery_replica_count value ignored with v2.0.0
#### Overview of the Issue

When running v2.0.0, the int value of 'shared_image_gallery_replica_count' is ignored. Resulting image version always has a replica count of 1.

Reverting to version 1.4.5 resolves the issue.

#### Reproduction Steps

```
packer {
  required_plugins {
    azure = {
      version = ">= 2.0.0"
      source  = "github.com/hashicorp/azure"
    }
  }
}
```

### Plugin and Packer version

2.0.0 and 1.9.4

```
variable "gallery_destination_rg"           {}
variable "gallery_image_name"               {}
variable "gallery_image_version"            {}
variable "gallery_name"                     {}
variable "gallery_subscription"             {}
variable "managed_image_name"               {}
variable "managed_image_resource_group"     {}
variable "tags"                             {}
variable "temp_compute_name"                {}
variable "tenant_id"                        {}
variable "user_assigned_managed_identities" {}
variable "virtual_network_name"             {}
variable "virtual_network_subnet_name"      {}

source "azure-arm" "win11" {
    azure_tags                        = var.tags
    tenant_id                         = var.tenant_id
    use_azure_cli_auth                = true
    user_assigned_managed_identities  = var.user_assigned_managed_identities
    image_offer                       = "office-365"
    image_publisher                   = "MicrosoftWindowsDesktop"
    image_sku                         = "win11-22h2-avd-m365"
    location                          = "uksouth"
    os_type                           = "Windows"
    vm_size                           = "Standard_D2_v4"

        gallery_name                  = var.gallery_name
        image_name                    = var.gallery_image_name
        image_version                 = var.gallery_image_version
        replication_regions           = ["uksouth"]
        resource_group                = var.gallery_destination_rg
        storage_account_type          = "Standard_LRS"
        subscription                  = var.gallery_subscription
    }
    virtual_network_name                              = var.virtual_network_name
    virtual_network_subnet_name                       = var.virtual_network_subnet_name
    shared_image_gallery_replica_count                = 3
    shared_gallery_image_version_end_of_life_date     = timeadd(timestamp(), "1440h")
    shared_gallery_image_version_exclude_from_latest  = false
}

packer {
  required_plugins {
    windows-update = {
      version = ">= 0.14.3"
      source  = "github.com/rgl/windows-update"
    }
    azure = {
      version = ">= 1.4.1"
      source  = "github.com/hashicorp/azure"
    }
  }
}

build {
    sources    = ["source.azure-arm.win11"]
    provisioner "windows-shell" {
        inline = ["mkdir c:\\test"]
    }
    provisioner "file" {
        source  = "../../scripts"
        destination = "c:/test"
    }
}
```

Windows 11 multi-session

Unable to supply.

JenGoldstrich:
Adds flag to allow SIG builds to use shallow replication https://learn.microsoft.com/en-us/azure/virtual-machines/shared-image-galleries?tabs=azure-cli#shallow-replication

Fix a bug where ReplicaCount was cast as int32 then an int64 leading to the wrong value being sent to the API

Closes #313 
Closes #341

### DIFF
--- a/.web-docs/components/builder/arm/README.md
+++ b/.web-docs/components/builder/arm/README.md
@@ -229,7 +229,7 @@ Providing `temp_resource_group_name` or `location` in combination with
 
 - `shared_image_gallery_replica_count` (int64) - The number of replicas of the Image Version to be created per region.
   Replica count must be between 1 and 100, but 50 replicas should be sufficient for most use cases.
-  This value can only be 1 when using shallow replication
+  When using shallow replication `use_shallow_replication=true` the value can only be 1.
 
 - `shared_gallery_image_version_exclude_from_latest` (bool) - If set to true, Virtual Machines deployed from the latest version of the
   Image Definition won't use this Image Version.
@@ -648,9 +648,7 @@ The shared_image_gallery_destination block is available for publishing a new ima
 
 - `specialized` (bool) - Set to true if publishing to a Specialized Gallery, this skips a call to set the build VM's OS state as Generalized
 
-- `use_shallow_replication` (bool) - Set to true to set the replication mode to shallow, which will publish the image version without replication.
-  This option results in a faster build but this image version's replication count and regions can not be updated after build when using shallow replication.
-  Setting a `shared_image_gallery_replica_count` or any `replication_regions` is unneccesary for shallow builds, as they can only be replicated to the build region, and must have a replica count of 1
+- `use_shallow_replication` (bool) - Setting a `shared_image_gallery_replica_count` or any `replication_regions` is unnecessary for shallow builds, as they can only replicate to the build region and must have a replica count of 1
   Refer to [Shallow Replication](https://learn.microsoft.com/en-us/azure/virtual-machines/shared-image-galleries?tabs=azure-cli#shallow-replication) for details on when to use shallow replication mode.
 
 <!-- End of code generated from the comments of the SharedImageGalleryDestination struct in builder/azure/arm/config.go; -->
--- a/.web-docs/components/builder/arm/README.md
+++ b/.web-docs/components/builder/arm/README.md
@@ -227,9 +227,9 @@ Providing `temp_resource_group_name` or `location` in combination with
 - `shared_gallery_image_version_end_of_life_date` (string) - The end of life date (2006-01-02T15:04:05.99Z) of the gallery Image Version. This property
   can be used for decommissioning purposes.
 
-- `shared_image_gallery_replica_count` (int32) - The number of replicas of the Image Version to be created per region. This
-  property would take effect for a region when regionalReplicaCount is not specified.
+- `shared_image_gallery_replica_count` (int64) - The number of replicas of the Image Version to be created per region.
   Replica count must be between 1 and 100, but 50 replicas should be sufficient for most use cases.
+  This value can only be 1 when using shallow replication
 
 - `shared_gallery_image_version_exclude_from_latest` (bool) - If set to true, Virtual Machines deployed from the latest version of the
   Image Definition won't use this Image Version.
@@ -335,8 +335,10 @@ Providing `temp_resource_group_name` or `location` in combination with
 
 - `build_resource_group_name` (string) - Specify an existing resource group to run the build in.
 
-- `build_key_vault_name` (string) - Specify an existing key vault to use for uploading certificates to the
+- `build_key_vault_name` (string) - Specify an existing key vault to use for uploading the certificate for the
   instance to connect.
+
+- `build_key_vault_secret_name` (string) - Specify the secret name to use for the certificate created in the key vault.
 
 - `build_key_vault_sku` (string) - Specify the KeyVault SKU to create during the build. Valid values are
   standard or premium. The default value is standard.
@@ -638,12 +640,18 @@ The shared_image_gallery_destination block is available for publishing a new ima
 
 - `image_version` (string) - Sig Destination Image Version
 
-- `replication_regions` ([]string) - Sig Destination Replication Regions
+- `replication_regions` ([]string) - A list of regions to replicate the image version in, by default the build location will be used as a replication region (the build location is either set in the location field, or the location of the resource group used in `build_resource_group_name` will be included.
+  Can not contain any region but the build region when using shallow replication
 
 - `storage_account_type` (string) - Specify a storage account type for the Shared Image Gallery Image Version.
   Defaults to `Standard_LRS`. Accepted values are `Standard_LRS`, `Standard_ZRS` and `Premium_LRS`
 
 - `specialized` (bool) - Set to true if publishing to a Specialized Gallery, this skips a call to set the build VM's OS state as Generalized
+
+- `use_shallow_replication` (bool) - Set to true to set the replication mode to shallow, which will publish the image version without replication.
+  This option results in a faster build but this image version's replication count and regions can not be updated after build when using shallow replication.
+  Setting a `shared_image_gallery_replica_count` or any `replication_regions` is unneccesary for shallow builds, as they can only be replicated to the build region, and must have a replica count of 1
+  Refer to [Shallow Replication](https://learn.microsoft.com/en-us/azure/virtual-machines/shared-image-galleries?tabs=azure-cli#shallow-replication) for details on when to use shallow replication mode.
 
 <!-- End of code generated from the comments of the SharedImageGalleryDestination struct in builder/azure/arm/config.go; -->
 
--- a/builder/azure/arm/builder.go
+++ b/builder/azure/arm/builder.go
@@ -234,8 +234,8 @@ func (b *Builder) Run(ctx context.Context, ui packersdk.Ui, hook packersdk.Hook)
 		// So we have to do this validation afterwards
 		// We should remove this logic builder and handle this logic via the `Step` pattern
 		if b.config.SharedGalleryDestination.SigDestinationUseShallowReplicationMode {
-			if len(normalizedReplicationRegions) != 1 || !foundMandatoryReplicationRegion {
-				return nil, fmt.Errorf("when `use_shallow_replication` is enabled the value of `replicated_regions` must match the build region specified by `location` or match the region of`build_resource_group_name`. ")
+			if len(b.config.SharedGalleryDestination.SigDestinationReplicationRegions) != 1 {
+				return nil, fmt.Errorf("when `use_shallow_replication` is enabled the value of `replicated_regions` must match the build region specified by `location` or match the region of `build_resource_group_name`.")
 			}
 		}
 		b.stateBag.Put(constants.ArmManagedImageSharedGalleryReplicationRegions, b.config.SharedGalleryDestination.SigDestinationReplicationRegions)
--- a/builder/azure/arm/builder.go
+++ b/builder/azure/arm/builder.go
@@ -213,21 +213,30 @@ func (b *Builder) Run(ctx context.Context, ui packersdk.Ui, hook packersdk.Hook)
 			return nil, fmt.Errorf("a gallery image version for image name:version %s:%s already exists in gallery %s", b.config.SharedGalleryDestination.SigDestinationImageName, b.config.SharedGalleryDestination.SigDestinationImageVersion, b.config.SharedGalleryDestination.SigDestinationGalleryName)
 		}
 
-		// SIG requires that replication regions include the region in which the Managed Image resides
-		managedImageLocation := normalizeAzureRegion(b.stateBag.Get(constants.ArmLocation).(string))
+		// SIG requires that replication regions include the region in which the created image version resides
+		buildLocation := normalizeAzureRegion(b.stateBag.Get(constants.ArmLocation).(string))
 		foundMandatoryReplicationRegion := false
 		var normalizedReplicationRegions []string
 		for _, region := range b.config.SharedGalleryDestination.SigDestinationReplicationRegions {
 			// change region to lower-case and strip spaces
 			normalizedRegion := normalizeAzureRegion(region)
 			normalizedReplicationRegions = append(normalizedReplicationRegions, normalizedRegion)
-			if strings.EqualFold(normalizedRegion, managedImageLocation) {
+			if strings.EqualFold(normalizedRegion, buildLocation) {
 				foundMandatoryReplicationRegion = true
 				continue
 			}
 		}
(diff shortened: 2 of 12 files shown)


maxilampert:
Maybe a config check should be added that source image/vm and image gallery replication destination region must be the same otherwise publishing will fail with an error?

builder/azure/arm/config.go
```diff
@@ -215,7 +217,7 @@ type Config struct {
 	// The number of replicas of the Image Version to be created per region. This
 	// property would take effect for a region when regionalReplicaCount is not specified.
 	// Replica count must be between 1 and 100, but 50 replicas should be sufficient for most use cases.
-	SharedGalleryImageVersionReplicaCount int32 `mapstructure:"shared_image_gallery_replica_count" required:"false"`
+	SharedGalleryImageVersionReplicaCount int64 `mapstructure:"shared_image_gallery_replica_count" required:"false"`
```
JenGoldstrich, line 220: In the new SDK this value is an int64, because this was being parsed as an int32, stored in the state bag, and then in `step_publish_to_shared_image_gallery` it was retrieved as an int64, this int would be incorrectly registered, leading to https://github.com/hashicorp/packer-plugin-azure/issues/341

builder/azure/arm/testdata/arm_linux_specialized.pkr.hcl
```diff
@@ -4,25 +4,26 @@ locals { timestamp = regex_replace(timestamp(), "[- TZ:]", "") }
```
JenGoldstrich, line 4: Ran packer fmt on these two templates after changing fields

builder/azure/arm/step_publish_to_shared_image_gallery.go
```diff
@@ -16,12 +16,26 @@ import (
 
 type StepPublishToSharedImageGallery struct {
 	client  *AzureClient
-	publish func(ctx context.Context, subscriptionID string, sourceID string, sharedImageGallery SharedImageGalleryDestination, miSGImageVersionEndOfLifeDate string, miSGImageVersionExcludeFromLatest bool, miSigReplicaCount int64, location string, diskEncryptionSetId string, tags map[string]string) (string, error)
+	publish func(publishArgs PublishArgs) (string, error)
```
JenGoldstrich, line 19: Refactored this to make this easier to unit test, and to read as there was a lot of parameters.

JenGoldstrich:
> Maybe a config check should be added that source image/vm and image gallery replication destination region must be the same otherwise publishing will fail with an error?

@maxilampert I agree that sounds like a good change, I think we should add it in a different PR though as this already changes a few things.

builder/azure/arm/step_publish_to_shared_image_gallery.go
nywilken: Nice way to consolidate. I would suggest breaking out the ctx seeing as it is not really a publishing arg and unexported. 
```suggestion
	publish func(ctx context.Context, args PublishArgs) (string, error)
```

builder/azure/arm/step_publish_to_shared_image_gallery.go
```diff
@@ -72,58 +86,60 @@ func getSigDestination(state multistep.StateBag) SharedImageGalleryDestination {
 	}
 }
 
-func (s *StepPublishToSharedImageGallery) publishToSig(ctx context.Context, subscriptionID string, sourceID string, sharedImageGallery SharedImageGalleryDestination, miSGImageVersionEndOfLifeDate string, miSGImageVersionExcludeFromLatest bool, miSigReplicaCount int64, location string, diskEncryptionSetId string, tags map[string]string) (string, error) {
-	replicationRegions := make([]galleryimageversions.TargetRegion, len(sharedImageGallery.SigDestinationReplicationRegions))
-	for i, v := range sharedImageGallery.SigDestinationReplicationRegions {
+func (s *StepPublishToSharedImageGallery) publishToSig(publishArgs PublishArgs) (string, error) {
```
nywilken, line 89: nit: The argument name can just be args since we know the type. `publishArgs PublishArgs` is a bit repetitive. 
```suggestion
func (s *StepPublishToSharedImageGallery) publishToSig(publishArgs PublishArgs) (string, error) {
```

docs-partials/builder/azure/arm/SharedImageGalleryDestination-not-required.mdx
```diff
@@ -17,4 +17,6 @@
 
 - `specialized` (bool) - Set to true if publishing to a Specialized Gallery, this skips a call to set the build VM's OS state as Generalized
 
+- `shallow_replication` (bool) - Set to true if you want to publish using Shallow Replication, this is faster but skips copying the full image to the gallery image region, read more about Shallow Replication [here](https://learn.microsoft.com/en-us/azure/virtual-machines/shared-image-galleries?tabs=azure-cli#shallow-replication)
```
nywilken, line 20: So this is a little confusing. At first I thought shallow replication meant that it would not wait for the replication to complete. But in reading the documentation. By setting this to true the image version is created but no replication occurs, as this is meant for testing purposes. 

I see the CLI has a flag called `replication-mode` which can be set to shallow. Would it be more explicit to cal call this attribute `skip_replication` or `use_shallow_replication`? 

```suggestion
- `use_shallow_replication` (bool) - Set to true to set the replication mode to shallow, which will publish the image version without replication. This option results in a faster build but skips the copying of the full gallery image to regions defined in `replication_regions`. Refer to [Shallow Replication](https://learn.microsoft.com/en-us/azure/virtual-machines/shared-image-galleries?tabs=azure-cli#shallow-replication) for details on when to use shallow replication mode. 
```

nywilken:
> Maybe a config check should be added that source image/vm and image gallery replication destination region must be the same otherwise publishing will fail with an error?

@JenGoldstrich @maxilampert does this mean that in order for shallow replication to work the replication_regions attribute can not contain a region that is not the same as the source image?  If so, what is the difference between using shallow replication mode and not specifying a replication at all?

nywilken (changes requested):
This is looking good. I left a few suggestions and questions about the feature. I know there is a small bug fix rolled into this change which we can break out if needed.

builder/azure/arm/step_publish_to_shared_image_gallery.go
JenGoldstrich: Done in newest commit

JenGoldstrich:
@maxilampert so I don't think we'll need to validate that specifically because the resulting image version is determined by the build location, if the source image and the build location differ the build will fail during VM publishing 

As to the difference between Shallow replication & full replication to 1 region, the main difference is that images using shallow replication can not have their replication information updated after the fact, whereas full replicated images can always go update their replica counts/regions

.web-docs/components/builder/arm/README.md
```diff
@@ -227,9 +227,9 @@ Providing `temp_resource_group_name` or `location` in combination with
 - `shared_gallery_image_version_end_of_life_date` (string) - The end of life date (2006-01-02T15:04:05.99Z) of the gallery Image Version. This property
   can be used for decommissioning purposes.
 
-- `shared_image_gallery_replica_count` (int32) - The number of replicas of the Image Version to be created per region. This
-  property would take effect for a region when regionalReplicaCount is not specified.
+- `shared_image_gallery_replica_count` (int64) - The number of replicas of the Image Version to be created per region.
   Replica count must be between 1 and 100, but 50 replicas should be sufficient for most use cases.
+  This value can only be 1 when using shallow replication
```
nywilken, line 232:
```suggestion
  When using shallow replication `use_shallow_replication=true` this value must be set to 1.
```

builder/azure/arm/builder.go
```diff
@@ -213,21 +213,30 @@ func (b *Builder) Run(ctx context.Context, ui packersdk.Ui, hook packersdk.Hook)
...
-			b.config.SharedGalleryDestination.SigDestinationReplicationRegions = append(normalizedReplicationRegions, managedImageLocation)
+			b.config.SharedGalleryDestination.SigDestinationReplicationRegions = append(normalizedReplicationRegions, buildLocation)
+		}
+		// TODO It would be better if validation could be handled in a central location
+		// Currently we rely on the build Resource Group being queried if used to get the build location
+		// So we have to do this validation afterwards
+		// We should remove this logic builder and handle this logic via the `Step` pattern
+		if b.config.SharedGalleryDestination.SigDestinationUseShallowReplicationMode {
+			if len(normalizedReplicationRegions) != 1 || !foundMandatoryReplicationRegion {
+				return nil, fmt.Errorf("when `use_shallow_replication` is set the build region (from `location` or `build_resource_group_name`) can only be region set in `replicated_regions`")
```
nywilken, line 238:
```suggestion
				return nil, fmt.Errorf("when `use_shallow_replication` is enabled the value of `replicated_regions` must match the build region specified by `location` or match the region of`build_resource_group_name`. ")
```

builder/azure/arm/config.go
```diff
@@ -212,10 +219,10 @@ type Config struct {
 	// The end of life date (2006-01-02T15:04:05.99Z) of the gallery Image Version. This property
 	// can be used for decommissioning purposes.
 	SharedGalleryImageVersionEndOfLifeDate string `mapstructure:"shared_gallery_image_version_end_of_life_date" required:"false"`
-	// The number of replicas of the Image Version to be created per region. This
-	// property would take effect for a region when regionalReplicaCount is not specified.
+	// The number of replicas of the Image Version to be created per region.
 	// Replica count must be between 1 and 100, but 50 replicas should be sufficient for most use cases.
-	SharedGalleryImageVersionReplicaCount int32 `mapstructure:"shared_image_gallery_replica_count" required:"false"`
+	// This value can only be 1 when using shallow replication
```
nywilken, line 224:
```suggestion
	// When using shallow replication `use_shallow_replication=true` the value can only be 1.
```

builder/azure/arm/config.go
```diff
@@ -1256,6 +1263,13 @@ func assertRequiredParametersSet(c *Config, errs *packersdk.MultiError) {
 		if c.SharedGalleryDestination.SigDestinationSubscription == "" {
 			c.SharedGalleryDestination.SigDestinationSubscription = c.ClientConfig.SubscriptionID
 		}
+
+		if c.SharedGalleryDestination.SigDestinationUseShallowReplicationMode {
+			// When the user doesn't set the replica count it uses the int null value (0) so we have to check for that
+			if c.SharedGalleryImageVersionReplicaCount != 1 && c.SharedGalleryImageVersionReplicaCount != 0 {
+				errs = packersdk.MultiErrorAppend(errs, fmt.Errorf("When using shallow replication the replica count can only be 1, leaving this value unset will default to 1"))
+			}
+		}
```
nywilken, line 1272: I think we can replaced the comment by setting the count to 1 if not value is specified. 
```suggestion

		if c.SharedGalleryDestination.SigDestinationUseShallowReplicationMode && c.SharedGalleryImageVersionReplicaCount  == 0 {
		c.SharedGalleryImageVersionReplicaCount = 1
		}
			
		if c.SharedGalleryImageVersionReplicaCount > 1 {
				errs = packersdk.MultiErrorAppend(errs, fmt.Errorf("When using shallow replication the replica count can only be 1, leaving this value unset will default to 1"))
			}
		}
```

builder/azure/arm/config_test.go
```diff
@@ -1283,6 +1283,85 @@ func TestConfigShouldAcceptAbsentManagedImageButPresentSharedImageGalleryDestina
...
+	var c Config
+	_, err := c.Prepare(config, getPackerConfiguration())
+	if err == nil {
+		t.Fatalf("expected config to reject invalid replica count using shallow replication but it was accepted")
+	} else if !strings.Contains(err.Error(), expectedErrorMessage) {
+		t.Fatalf("unexpected rejection reason, expected %s to contain %s", err.Error(), expectedErrorMessage)
+	}
+}
+
+func TestConfigShouldAcceptShallowReplicationWithWithSetReplicaCount(t *testing.T) {
```
nywilken, line 1315:
```suggestion
func TestConfigShouldAcceptShallowReplicationWithReplicaCount(t *testing.T) {
```

builder/azure/arm/step_publish_to_shared_image_gallery.go
```diff
@@ -187,9 +202,28 @@ func (s *StepPublishToSharedImageGallery) Run(ctx context.Context, stateBag mult
 	s.say(fmt.Sprintf(" -> SIG image version endoflife date      : '%s'", miSGImageVersionEndOfLifeDate))
 	s.say(fmt.Sprintf(" -> SIG image version exclude from latest : '%t'", miSGImageVersionExcludeFromLatest))
 	s.say(fmt.Sprintf(" -> SIG replica count [1, 100]            : '%d'", miSigReplicaCount))
-
+	shallowReplicationMode := stateBag.Get(constants.ArmSharedImageGalleryDestinationShallowReplication).(bool)
+	replicationMode := galleryimageversions.ReplicationModeFull
```
nywilken, line 206: nit: ordering to avoid the blip in the middle. 
```suggestion
	replicationMode := galleryimageversions.ReplicationModeFull
	shallowReplicationMode := stateBag.Get(constants.ArmSharedImageGalleryDestinationShallowReplication).(bool)
```

nywilken (approved):
I have a few suggestions on the documentation and a few small nits but otherwise looks good. Approving in advance to unblock the merge when you've applied the suggestions.

.web-docs/components/builder/arm/README.md
JenGoldstrich: I will re-run make generate to create this suggestion from the config changes

PR #338, opened 2023-09-16
Change type of `encryption_at_host` to `*bool`

KMConner:
In v1.4.4, `encryption_at_host` flag was introduced, but we need to enable EncryptionAtHost  on Azure subscrioptions in oder to use `encryption_at_host` feature.

`packer build` with arm builder always fails on Azure subscription if EncryptionAtHost feature is not enabled.
This error happens even if `encryption_at_host` is set to false or `encryption_at_host` is not specified.

The below is the error message.

```
[{"code":"InvalidParameter","message":"The property 'securityProfile.encryptionAtHost' is not valid because the 'Microsoft.Compute/EncryptionAtHost' feature is not enabled for this subscription.","target":"securityProfile.encryptionAtHost"}]
```

This is because `securityProfile.encryptionAtHost` property is always specified in the template deployed while build.
The property should not be specified when EncryptionAtHost  feature is not enabled (Error happend even if `false` is specified).

So, I changed `encryption_at_host` argument from `bool` to `*bool`, so that `securityProfile.encryptionAtHost` property is not contained in the deployed template when `encryption_at_host` is not specified.

### DIFF
--- a/builder/azure/common/template/template_builder.go
+++ b/builder/azure/common/template/template_builder.go
@@ -512,7 +512,7 @@ func (s *TemplateBuilder) SetLicenseType(licenseType string) error {
 	return nil
 }
 
-func (s *TemplateBuilder) SetSecurityProfile(secureBootEnabled bool, vtpmEnabled bool, encryptionAtHost bool) error {
+func (s *TemplateBuilder) SetSecurityProfile(secureBootEnabled bool, vtpmEnabled bool, encryptionAtHost *bool) error {
 	resource, err := s.getResourceByType(resourceVirtualMachine)
 	if err != nil {
 		return err
@@ -526,7 +526,7 @@ func (s *TemplateBuilder) SetSecurityProfile(secureBootEnabled bool, vtpmEnabled
 		resource.Properties.SecurityProfile.UefiSettings.SecureBootEnabled = common.BoolPtr(secureBootEnabled)
 		resource.Properties.SecurityProfile.UefiSettings.VTpmEnabled = common.BoolPtr(vtpmEnabled)
 	}
-	resource.Properties.SecurityProfile.EncryptionAtHost = common.BoolPtr(encryptionAtHost)
+	resource.Properties.SecurityProfile.EncryptionAtHost = encryptionAtHost
 
 	return nil
 }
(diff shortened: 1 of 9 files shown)


builder/azure/common/template/template_builder.go
```diff
@@ -533,7 +533,7 @@ func (s *TemplateBuilder) SetSecurityProfile(secureBootEnabled bool, vtpmEnabled
 		resource.Properties.SecurityProfile.UefiSettings.SecureBootEnabled = common.BoolPtr(secureBootEnabled)
 		resource.Properties.SecurityProfile.UefiSettings.VTpmEnabled = common.BoolPtr(vtpmEnabled)
 	}
-	resource.Properties.SecurityProfile.EncryptionAtHost = common.BoolPtr(encryptionAtHost)
+	resource.Properties.SecurityProfile.EncryptionAtHost = encryptionAtHost
```
lbajolet-hashicorp, line 536: I wonder, should we only set the parameter if true instead of always setting a value? This would be a different (simpler) implementation than what you propose here, that being said maybe there's a case for setting `encryptionAtHost` to `false` manually?
If so the current approach makes sense and we can accept it as-is.

cc @JenGoldstrich since you know Azure much better than I do

lbajolet-hashicorp (commented):
Hi @KMConner,

Thanks for submitting this PR, I reviewed the code, and while this works, I wonder if we can only add the attribute to the template if it is set, and not systematically.
This may not cover all cases though, but I'll leave this suggestion here, feel free to respond when you have time.

Other than this concern, the code looks legit, I'll come back to this when there's an update, and if we decide to keep the code as-is, we can merge.

builder/azure/common/template/template_builder.go
JenGoldstrich: I'm conflicted tbh, to do what you suggest it would need to be refactored to be something like this
```
if encryptionAtHost != nil && *encryptionAtHost {
     	resource.Properties.SecurityProfile.EncryptionAtHost = encryptionAtHost
}
```

While we would avoid setting the the field in the template by changing to this, it requires that extra nil check and a true/false. check, when we never actually send false, only true and nil, since sending false is what causes this issue for Azure Subscriptions without this feature enabled (TBH this feels like an API bug but thats an aside). But I will accept either solution tbh

JenGoldstrich:
Hey @KMConner thanks for your PR, I merged a PR that modifies the template builder quite a lot, do you mind rebasing on top of those changes, and making sure your changes still resolve your issue? Honestly after bumping the deployment API version we might have resolved this issue so it'd be helpful for you to re-test with the version of the plugin on main (If you'd prefer I can trigger a release next week for you to try before we move forward).  If you still experience this issue on the new API I'm happy to merge this change to unblock you from upgrading your Azure plugin

JenGoldstrich (approved):
Thanks for rebasing, @KMConner this LGTM, I manually tested this field and ran acceptance tests, thanks for your contribution!

Issue #358, opened 2024-01-11
Packer logs contain adminPassword in plain text


jerryc3:
When filing a bug, please include the following headings if possible. Any
example text in this template can be deleted.

#### Overview of the Issue

AzureImageBuilder team got a customer complaint that packer logs contain VM adminPassword in plain text. Can packer redact this password from the logs?

#### Reproduction Steps

Steps to reproduce this issue

### Plugin and Packer version

From `packer version`

### Simplified Packer Buildfile

If the file is longer than a few dozen lines, please include the URL to the
[gist](https://gist.github.com/) of the log or use the [Github detailed
format](https://gist.github.com/ericclemmons/b146fe5da72ca1f706b2ef72a20ac39d)
instead of posting it directly in the issue.

### Operating system and Environment details

OS, Architecture, and any other information you can provide about the
environment.

### Log Fragments and crash.log files

What customer saw:

Azure request method="POST" request="[https://management.azure.com/subscriptions/96ff5940-d6af-45ee-ae2b-3451cc65a2b5/resourceGroups/IT_myrg_testImageTemplate1_e73c9f6f-0025-435f-be0f-d0553984fc17/providers/Microsoft.Resources/deployments/pkrdp3b3qlyzo8a/validate?[REDACTED]"](https://management.azure.com/subscriptions/96ff5940-d6af-45ee-ae2b-3451cc65a2b5/resourceGroups/IT_myrg_testImageTemplate1_e73c9f6f-0025-435f-be0f-d0553984fc17/providers/Microsoft.Resources/deployments/pkrdp3b3qlyzo8a/validate?[REDACTED]%22) body="{\"properties\":{\"mode\":\"Incremental\",\"parameters\":{\"adminPassword\":{\"value\":\"[HIDDEN MANUALLY]\"},

[...]

PR #365, opened 2024-01-23
log: add mini log library

lbajolet-hashicorp:
The Azure log library mimicks what the standard log library allows (with fewer functions exposed), and uses what the Packer SDK exposes for filtering out sensitive strings.

We have to do this here for censoring potential sensitive data separately from Packer itself, as the secrets produced/consumed by the plugin itself aren't necessarily known by Packer, so it cannot censor them from its output, and we have no way for a plugin to declare what is sensitive in its logs.

Therefore, we do it at log-time, in the plugin process, so Packer receives pre-censored logs, which it can render normally through its pipeline.

Then we adopt it for all our logging uses in the plugin, which should censor all the strings we want in the plugin process.

Closes: #358


nywilken (approved):
Nice approach on wrapping the logger any reason why there couldn't be a Set method on the common log that wrap `packer.LogSecretFilter.Set(secret)`?

This is a question not a blocker to merging

lbajolet-hashicorp:
Honestly that is a good question, I have not considered it since we have access to the `LogSecretFilter` already, and there were some use of that in the code (the `adminPassword` was already registering through this path).
This could be a good idea to add this just for better visibility.

PR #368, opened 2024-02-01
builder: don't log requests/responses by default

lbajolet-hashicorp:
The logs of the Azure builder are very verbose by default, as they log the contents of every request/response to/from the Azure APIs.

For regular debugging of a template, this is not necessary, as just looking at the errors returned by the calls will generally give enough information to proceed with fixing the template.

These logs may be relevant when attempting to debug the plugin itself however, so we add one extra environment variable to allow the plugin to print them out, PACKER_AZURE_DEBUG_LOGS.
If it is defined, and non empty, the requests and responses will be dumped to stderr, otherwise, they are silent.

### DIFF
--- a/builder/azure/common/env.go
+++ b/builder/azure/common/env.go
@@ -1,0 +1,14 @@
+package common
+
+import "os"
+
+const AzureDebugLogsEnvVar string = "PACKER_AZURE_DEBUG_LOG"
+
+func IsDebugEnabled() bool {
+	debug, defined := os.LookupEnv(AzureDebugLogsEnvVar)
+	if !defined {
+		return false
+	}
+
+	return debug != ""
+}
--- a/docs/README.md
+++ b/docs/README.md
@@ -87,3 +87,13 @@ you should specify `subscription_id`, `client_id` and one of `client_secret`,
   associated with your service principal principal. See [Azure Active
   Directory docs](https://docs.microsoft.com/en-us/azure/active-directory/develop/active-directory-certificate-credentials)
   for more information.
+
+## Troubleshooting
+
+As with any Packer plugin, you may produce verbose logs to troubleshoot if the default output does not help narrow down the issue.
+To do so, you can set the `PACKER_LOG` environment variable to a non-zero value (e.g. "1" or any non-empty string) to enable verbose logs that can help you figure out which part of the process errors, and hopefully why.
+
+In addition to this, you may also enable the `PACKER_AZURE_DEBUG_LOG` environment variable alongside `PACKER_LOG` for visibility into the API calls being made to the Azure platform.
+Enabling this will add HTTP response inspection in the logs and the body sent with each request to the Azure APIs.
+
+~> Warning: the `PACKER_AZURE_DEBUG_LOG` variable contains a high degree of verbosity and may expose sensitive information in the logs. For this reason, we strongly advise only enabling this in a trusted environment and only for a temporary debugging session.
--- a/.web-docs/README.md
+++ b/.web-docs/README.md
@@ -100,3 +100,13 @@ you should specify `subscription_id`, `client_id` and one of `client_secret`,
   associated with your service principal principal. See [Azure Active
   Directory docs](https://docs.microsoft.com/en-us/azure/active-directory/develop/active-directory-certificate-credentials)
   for more information.
+
+## Troubleshooting
+
+As with any Packer plugin, you may produce verbose logs to troubleshoot if the default output does not help narrow down the issue.
+To do so, you can set the `PACKER_LOG` environment variable to a non-zero value (e.g. "1" or any non-empty string) to enable verbose logs that can help you figure out which part of the process errors, and hopefully why.
+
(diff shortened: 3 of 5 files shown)


builder/azure/common/inspector.go
```diff
@@ -0,0 +1,109 @@
+// Copyright (c) HashiCorp, Inc.
+// SPDX-License-Identifier: MPL-2.0
+
+package common
+
+import (
```
nywilken, line 6: Is this common file meant to replace the other two inspectors? 

I see the code added to each builder but wonder if they can be consolidated.

builder/azure/common/env.go
```diff
@@ -0,0 +1,14 @@
+package common
+
+import "os"
+
+const DebugLogsEnvVar string = "PACKER_AZURE_DEBUG_LOGS"
```
nywilken, line 5: nit: Prefix with Azure to denote these are Azure debug log and drop the `s` for consistency with _LOG
```suggestion
const AzureDebugLogsEnvVar string = "PACKER_AZURE_DEBUG_LOG"
```

nywilken (commented):
I think the approach is sound and I see why we want to keep and not just get rid of them. Where should this information be documented?

builder/azure/common/inspector.go
lbajolet-hashicorp: Oh, nevermind this, I had tried to reify the inspector implementation in common, but since the private functions like `handleBody` are used in different places in each builder, I opted no to do this right away.
I'll remove this from the PR, we can revisit later.

lbajolet-hashicorp:
> I think the approach is sound and I see why we want to keep and not just get rid of them. Where should this information be documented?

That's a good question. This is probably something we should detail in the main `README` for the plugin.
Or we could leave it undocumented and only share about it if needed; I'm not sure this is the kind of capability we want to make every user aware of because of the risk of dumping that traffic on stderr.
What do you think?

nywilken:
> Or we could leave it undocumented and only share about it if needed; I'm not sure this is the kind of capability we want to make every user aware of because of the risk of dumping that traffic on stderr.

I don't think leaving it undocumented helps the situation. I would rather inform users of the potential risk as opposed to having say I didn't know that was possible. In this case, I think the right thing to do is add a section to the integration README for Troubleshooting. 

In this section, we can explain the use of `PACKER_LOG` and `PACKER_AZURE_DEBUG_LOG` for gaining greater visibility into the API calls made by the plugin. 

Setting `PACKER_AZURE_DEBUG_LOG` will enable logging of all HTTP requests made to Azure API endpoints to stdout, which may contain sensitive information such as passwords or access keys as raw text. As a security precaution, Azure debug logging is not available by default. Users wishing to enable such granular logging are encouraged to do so on an as-needed basis and only on trusted hosts.

lbajolet-hashicorp:
@nywilken good call; I like the `Troubleshooting` section idea.

I've just pushed docs update for this, when you have time please let me know what you think about it.

docs/README.md
```diff
@@ -87,3 +87,15 @@ you should specify `subscription_id`, `client_id` and one of `client_secret`,
...
+
+As with any Packer plugin, you may produce verbose logs if you encounter a problem, and the normal logs don't help narrow down the issue.
+To do so, you will need to set the `PACKER_LOG` environment variable to a non-zero value (e.g. "1", "y", or any non-empty string).
+This will enable some verbose logs that can help you figure out which part of the process errors, and hopefully why.
+
+In addition to this, you may also enable the `PACKER_AZURE_DEBUG_LOG` environment variable, in addition to `PACKER_LOG`.
+Enabling this will add request/response inspection in the logs, along with the body sent with each request to the Azure APIs.
+
+~> Warning: the `PACKER_AZURE_DEBUG_LOG` variable is very verbose, and may expose sensitive information in the logs.
+   For this reason, we strongly advise only enabling this in a trusted environment, and only for a temporary debugging session.
```
nywilken, line 101: Small rewrite to remove wordy sentences 

```suggestion

As with any Packer plugin, you may produce verbose logs to troubleshoot if the default output does not help narrow down the issue. To do so, you can set the `PACKER_LOG` environment variable to a non-zero value (e.g. "1" or any non-empty string) to enable verbose logs that can help you figure out which part of the process errors and hopefully why.

In addition to this, you may also enable the `PACKER_AZURE_DEBUG_LOG` environment variable in alongside  `PACKER_LOG` for visibility into the API calls being made to the Azure platform. Enabling this will add HTTP response inspection in the logs and the body sent with each request to the Azure APIs.

~> Warning: the `PACKER_AZURE_DEBUG_LOG` variable contains a high degree of verbosity and may expose sensitive information in the logs. For this reason, we strongly advise only enabling this in a trusted environment and only for a temporary debugging session.
```

nywilken (approved):
Small changes to the text but otherwise good to go.

docs/README.md
lbajolet-hashicorp: LGTM, thanks for the reword

Issue #264, opened 2023-01-31
Disk Encryption Set ID does not work for Replicated Regions


JenGoldstrich:
I realized after releasing v1.4.0 that setting a `disk_encryption_set_id` will not work when replicating a SIG image to multiple regions, since Disk Encryption Set's can only exist in one region if you're replicating a shared image definition to a different region, that DES won't exist in that region, this should be fixed adding an extra field to shared_image_gallery_destination, for these replication regions encryption IDs, it should take in an array of strings that are at the same index of the replication region, something like this in HCL

```
shared_image_gallery_destination {
     resource_group = "test"
     gallery_name = "gallery"
     image_name = "image"
     image_version = "1.0.0"
     replication_regions  = ["East US", "East US 2", "East US 3"]
     replication_region_disk_encryption_set_ids = ["encryption-for-east-us-id", "encryption-for-east-us-id-2", "encryption-for-east-us-3-id"]
}
```

markti:
this does in fact look like a bug... as disk encryption sets are regional, it looks like you need to provide a different one for each region you replicate to. 

The code here seems to be the problem:

https://github.com/hashicorp/packer-plugin-azure/blob/10a55d5fbd83bf263dc3d6f2a6ab8cd4d7f2a1b7/builder/azure/arm/step_publish_to_shared_image_gallery.go#L33

the attribute "disk_encryption_set_id" is mapped to variable "DiskEncryptionSetId" but it is set at the builder level and not specified for each geo-replicated region.

darrens280:
Adding support for `DiskEncryptionSetId` in Azure Plugin v2.0.3 (released last night), using new syntax of the `target_region` (with one replicated region) block to replace the deprecated `replicated_regions` attribute inside `shared_image_gallery_destination`, this breaks the pipeline, at the point where the Image is being captured into the Compute Gallery, with the error below. We don't use Disk Encryption Sets, but it seems to be mandatory (?)

```
ERROR: -> LinkedInvalidPropertyId : Property id '' at path 'properties.publishingProfile.targetRegions[0].encryption.osDiskImage.diskEncryptionSetId' is invalid. Expect fully qualified resource Id that start with '/subscriptions/{subscriptionId}' or '/providers/{resourceProviderNamespace}/'
```

When I revert and use Azure Plugin v2.0.2, and `replicated_regions`, then all works fine.

Thanks
Darren

nywilken:
[...]

PR #371, opened 2024-02-05
[HPR-1606] Add support specifying disk encryption set keys across replicated regions

nywilken:
Add support for target_region block in SharedImageGalleryDestination

As an alternative to specifying replication_regions for a Shared Image
Gallery Destination, users can not use one or more `target_region` block
attributes to define the target regions for a shared image version. The
target_region block attribute supports the setting of individual disk
encryption key id for multi-region replication of encrypted
disks using CMKs.

The `target_region` attribute is meant to replace the now deprecated
`replication_regions` attribute. Internally, any values set for
replication_regions will be transferred into a set of target_region
blocks to provide the same level of functionality. The one difference being that
Packer will no longer add a default entry for the SIG build location
when using target_region blocks.

Closes: https://github.com/hashicorp/packer-plugin-azure/issues/264

#### Example Config
```hcl
variable "resource_group" {
  type    = string
  default = "${env("ARM_RESOURCE_GROUP)}"
}

source "azure-arm" "autogenerated_1" {
  use_azure_cli_auth                = true
  image_offer                       = "UbuntuServer"
  image_publisher                   = "Canonical"
  image_sku                         = "16.04-LTS"
  location                          = "East US"
  os_type                           = "Linux"
  vm_size         = "Standard_DS2_v2"
    gallery_name        = "PackerSigGallery"
    image_name          = "PackerSigImageDefinition"
    image_version       = "1.0.1"
    resource_group      = var.resource_group
    target_region {
        name = "West US 2"
        disk_encryption_set_id = "KEY"
    }
    target_region {
        name = "East US"
        disk_encryption_set_id = "KEY"
    }
  }
}

build {
  sources = ["source.azure-arm.autogenerated_1"]

  provisioner "shell" {
    execute_command = "chmod +x {{ .Path }}; {{ .Vars }} sudo -E sh '{{ .Path }}'"
    inline          = ["apt-get update", "apt-get upgrade -y", "/usr/sbin/waagent -force -deprovision+user && export HISTSIZE=0 && sync"]
    inline_shebang  = "/bin/sh -x"
  }

}
```

### DIFF
--- a/builder/azure/arm/builder.go
+++ b/builder/azure/arm/builder.go
@@ -215,20 +215,6 @@ func (b *Builder) Run(ctx context.Context, ui packersdk.Ui, hook packersdk.Hook)
 			return nil, fmt.Errorf("a gallery image version for image name:version %s:%s already exists in gallery %s", b.config.SharedGalleryDestination.SigDestinationImageName, b.config.SharedGalleryDestination.SigDestinationImageVersion, b.config.SharedGalleryDestination.SigDestinationGalleryName)
 		}
 
-		// TODO It would be better if validation could be handled in a central location
-		// Currently we rely on the build Resource Group being queried if used to get the build location
-		// So we have to do this validation afterwards
-		// We should remove this logic builder and handle this logic via the `Step` pattern
-		if b.config.SharedGalleryDestination.SigDestinationUseShallowReplicationMode {
-			err := errors.New("when `use_shallow_replication` there can only be one destination region defined and it must match `location` or match the region of `build_resource_group_name`.")
-			switch {
-			case len(b.config.SharedGalleryDestination.SigDestinationTargetRegions) > 1:
-				return nil, err
-			case len(b.config.SharedGalleryDestination.SigDestinationReplicationRegions) > 1:
-				return nil, err
-			}
-		}
-
 		if len(b.config.SharedGalleryDestination.SigDestinationTargetRegions) > 0 {
 			normalizedRegions := make([]TargetRegion, 0, len(b.config.SharedGalleryDestination.SigDestinationTargetRegions))
 			for _, tr := range b.config.SharedGalleryDestination.SigDestinationTargetRegions {
--- a/builder/azure/arm/builder.go
+++ b/builder/azure/arm/builder.go
@@ -215,11 +215,6 @@ func (b *Builder) Run(ctx context.Context, ui packersdk.Ui, hook packersdk.Hook)
 			return nil, fmt.Errorf("a gallery image version for image name:version %s:%s already exists in gallery %s", b.config.SharedGalleryDestination.SigDestinationImageName, b.config.SharedGalleryDestination.SigDestinationImageVersion, b.config.SharedGalleryDestination.SigDestinationGalleryName)
 		}
 
-		// Validate target region settings; it can be the deprecated replicated_regsions attribute or multiple target_region blocks
-		if (len(b.config.SharedGalleryDestination.SigDestinationReplicationRegions) > 0) && (len(b.config.SharedGalleryDestination.SigDestinationTargetRegions) > 0) {
-			return nil, fmt.Errorf("`replicated_regions` can not be defined alongside `target_regions`; you can defined a target_region for each destination region you wish to replicate to.")
-		}
-
 		// TODO It would be better if validation could be handled in a central location
 		// Currently we rely on the build Resource Group being queried if used to get the build location
 		// So we have to do this validation afterwards
--- a/builder/azure/arm/builder.go
+++ b/builder/azure/arm/builder.go
@@ -215,33 +215,64 @@ func (b *Builder) Run(ctx context.Context, ui packersdk.Ui, hook packersdk.Hook)
 			return nil, fmt.Errorf("a gallery image version for image name:version %s:%s already exists in gallery %s", b.config.SharedGalleryDestination.SigDestinationImageName, b.config.SharedGalleryDestination.SigDestinationImageVersion, b.config.SharedGalleryDestination.SigDestinationGalleryName)
 		}
 
-		// SIG requires that replication regions include the region in which the created image version resides
-		buildLocation := normalizeAzureRegion(b.stateBag.Get(constants.ArmLocation).(string))
-		foundMandatoryReplicationRegion := false
-		var normalizedReplicationRegions []string
-		for _, region := range b.config.SharedGalleryDestination.SigDestinationReplicationRegions {
-			// change region to lower-case and strip spaces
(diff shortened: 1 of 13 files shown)


builder/azure/arm/builder.go
```diff
@@ -213,33 +213,64 @@ func (b *Builder) Run(ctx context.Context, ui packersdk.Ui, hook packersdk.Hook)
...
-			normalizedRegion := normalizeAzureRegion(region)
-			normalizedReplicationRegions = append(normalizedReplicationRegions, normalizedRegion)
-			if strings.EqualFold(normalizedRegion, buildLocation) {
-				foundMandatoryReplicationRegion = true
-				continue
-			}
-		}
-		if foundMandatoryReplicationRegion == false {
-			b.config.SharedGalleryDestination.SigDestinationReplicationRegions = append(normalizedReplicationRegions, buildLocation)
+		// Validate target region settings; it can be the deprecated replicated_regsions attribute or multiple target_region blocks
```
nywilken, line 216:
```suggestion
		// Validate target region settings; it can be the deprecated replicated_regions attribute or multiple target_region blocks
```

builder/azure/arm/builder.go
```diff
@@ -213,33 +213,64 @@ func (b *Builder) Run(ctx context.Context, ui packersdk.Ui, hook packersdk.Hook)
...
+		if (len(b.config.SharedGalleryDestination.SigDestinationReplicationRegions) > 0) && (len(b.config.SharedGalleryDestination.SigDestinationTargetRegions) > 0) {
+			return nil, fmt.Errorf("`replicated_regions` can not be defined alongside `target_regions`; you can defined a target_region for each destination region you wish to replicate to.")
```
nywilken, line 218:
```suggestion
			return nil, fmt.Errorf("`replicated_regions` can not be defined alongside `target_region`; you can defined a target_region for each destination region you wish to replicate to.")
```

builder/azure/arm/builder.go
```diff
@@ -213,33 +213,64 @@ func (b *Builder) Run(ctx context.Context, ui packersdk.Ui, hook packersdk.Hook)
...
 		}
+
 		// TODO It would be better if validation could be handled in a central location
```
JenGoldstrich, line 221: This code looks good but I wonder if we should add this validation to its own function or its own step, that way we could more easily unit test this logic in isolation, feel free to take that feedback or leave it, but I think it'd be good to add some sort of unit test coverage to ensure that we don't break the backwards compatibility you've established here in future updates to this code

builder/azure/arm/builder.go
```diff
@@ -213,33 +213,64 @@ func (b *Builder) Run(ctx context.Context, ui packersdk.Ui, hook packersdk.Hook)
...
-			normalizedReplicationRegions = append(normalizedReplicationRegions, normalizedRegion)
-			if strings.EqualFold(normalizedRegion, buildLocation) {
-				foundMandatoryReplicationRegion = true
-				continue
-			}
-		}
-		if foundMandatoryReplicationRegion == false {
-			b.config.SharedGalleryDestination.SigDestinationReplicationRegions = append(normalizedReplicationRegions, buildLocation)
+		// Validate target region settings; it can be the deprecated replicated_regsions attribute or multiple target_region blocks
+		if (len(b.config.SharedGalleryDestination.SigDestinationReplicationRegions) > 0) && (len(b.config.SharedGalleryDestination.SigDestinationTargetRegions) > 0) {
```
JenGoldstrich, line 217: Could we check for this before we compute the regions? Like in the config validation we could just check if `target_regions` and `replicated_regions` are set together, that way we fail a bit earlier when the config is invalid, what do you think?

nywilken: Yeah we can do that. This checks seems like it is  bit late!

nywilken: Done

builder/azure/arm/builder.go
nywilken: I moved this into config.go and added a few tests for it. Let me know what you think.

builder/azure/arm/config_test.go
```diff
@@ -1362,6 +1362,116 @@ func TestConfigShouldAcceptShallowReplicationWithWithUnsetReplicaCount(t *testin
 		t.Fatalf("expected config to accept shallow replication with unset replica count build: %v", err)
 	}
 }
+
+func TestConfigValidateShallowReplicationRegion(t *testing.T) {
```
JenGoldstrich, line 1366: Awesome! thanks for adding these

builder/azure/arm/builder.go
```diff
@@ -213,33 +213,45 @@ func (b *Builder) Run(ctx context.Context, ui packersdk.Ui, hook packersdk.Hook)
...
-		// TODO It would be better if validation could be handled in a central location
-		// Currently we rely on the build Resource Group being queried if used to get the build location
-		// So we have to do this validation afterwards
-		// We should remove this logic builder and handle this logic via the `Step` pattern
-		if b.config.SharedGalleryDestination.SigDestinationUseShallowReplicationMode {
-			if len(b.config.SharedGalleryDestination.SigDestinationReplicationRegions) != 1 {
-				return nil, fmt.Errorf("when `use_shallow_replication` is enabled the value of `replicated_regions` must match the build region specified by `location` or match the region of `build_resource_group_name`.")
+
+		// Convert deprecated replication_regions to []TargetRegion
+		if len(b.config.SharedGalleryDestination.SigDestinationReplicationRegions) > 0 {
```
JenGoldstrich, line 226: Not blocking and I think this could be done in a follow up PR but I think we should move this block as well, into a new function called ConvertReplicatedRegions in can still exist in the builder but that way we can unit test it in isolation so that we don't break backwards compatibility in the future with this unintentionally.

JenGoldstrich (approved):
This largely LGTM, nice work on this issue Wilken, I left a thought about a block in `builder.go` I think we could break off into a function and unit test in isolation, but I'm happy to do that refactoring in a follow up PR if you'd prefer, lemme know what you think!

builder/azure/arm/builder.go
nywilken: So I agree this could help with testing but there is a bunch of nuance logic happening there that I wouldn't want to hide in a function. Breaking out the for loop is straight forward but  we would run it after normalization. There are a few tests already that if given replication_regions asserts that the the correct number of target_regions are crated. If that helps.

JenGoldstrich: That makes sense, thanks for clarifying!

nywilken: I thought about how to move this stuff into a step_* but 😵 - maybe after a day something comes up. Thanks for this suggestion.

PR #373, opened 2024-02-08
Add Security Type Option for Confidential and Trusted Launch VMs

### LINKED ISSUE #372: Support a `SecurityType` attribute in order to build ConfidentialVM images
#### Description

As mentioned in #290, the Azure Packer plugin should be enhanced to support ConfidentialVMs. 

Currently, VMs with a standard security type can be imaged and published to sig definitions with `SecurityType=ConfidentialVMSupported`. However, the plugin errors during a sig image publish when it attempts to publish to a sig definition requiring a ConfidentialVM, `SecurityType=ConfidentialVM`

```console
==> Confidential VM is not supported for the source id '/subscriptions/<guid>/resourceGroups/<rg_name>/providers/Microsoft.Compute/images/u22-cvm'. Currently Snapshot, Disk and VM sources are supported for the Confidential VM security type." Target="galleryImageVersion.properties.storageProfile.osDiskImage.source.id"
```

The requirements for a ConfidentialVM are as such:

1. vTPM is enabled --> `vtpm_enabled` in packer-plugin-azure
2. Secure Boot is enabled --> `secure_boot_enabled` in packer-plugin-azure
3. _**Security Type set to `ConfidentialVM --> not currently supported**_
4. Encrypting the OS Disk (PaaS key or CMK) --> PaaS key not currently support, CMK supported in packer-plugin-azure via `disk_encryption_set_id`

The next issue is that currently, the plugin assumes the `SecurityType` of the VM when `vtpm_enabled` and `secure_boot_enabled` are both set to `true`. The plugin assumes the `SecurityType` to be `TrustedLaunch`. 

https://github.com/hashicorp/packer-plugin-azure/blob/24449b4f2a98c7fd0f38b7473bcb079fdb81fe22/builder/azure/common/template/template_builder.go#L522

```console
==> azure-arm.linux_mktpl: polling after CreateOrUpdate: Code="DeploymentFailed" Message="At least one resource deployment operation failed. Please list deployment operations for details. Please see https://aka.ms/arm-deployment-operations for usage details." Target="/subscriptions/4a727e1d-6e1e-4b60-bf3e-bc5bc918dc25/resourceGroups/rg-eus-image-builds/providers/Microsoft.Resources/deployments/pkrdpp2lcl39ikf" Details=[{"code":"BadRequest","message":"The VM size 'Standard_DC2as_v5' is not supported for creation of VMs and Virtual Machine Scale Sets with 'TrustedLaunch' security type. For more information, see https://aka.ms/TrustedLaunch-FAQ"}]
```

What I am proposing, and have working code for, is to allow users to pass in a `security_type` attribute. As of now, it will be limited to simply `ConfidentialVM` and `TrustedLaunch` with the ability to further extend down the road. If users want to take advantage of confidential VMs they can do so by configuring the security type, vtpm, secure boot, and disk encryption settings.

```hcl
...
# Confidential VMs
vtpm_enabled = true
secure_boot_enabled = true
security_type = "ConfidentialVM"
disk_encryption_set_id = "<azure_disk_encryption_set_id>"
...
```

#### Use Case(s)

- Deploying images to Shared Image Definitions where the SecurityType is `ConfidentialVM` (mandated)
- Deploying the build VM as a ConfidentialVM from a ConfidentialVM-specific marketplace image offering to validate the image
```hcl
...
  image_publisher = "Canonical"
  image_offer     = "0001-com-ubuntu-confidential-vm-jammy"
  image_sku       = "22_04-lts-cvm"
...
```

#### Potential configuration

- Create a new plugin attribute to take in `security_type`
- Validate that the attribute is set to `ConfidentialVM` or `TrustedLaunch`
- Set the security type in the ARM template accordingly

As I mentioned above, I have the code written for this change and will submit a PR for it. 🙂

julian1059:
Closes #372 and answers #290 

Please let me know what you think of these changes. I am happy to make any adjustments needed.


julian1059:
I found an issue deploying a ConfidentialVM using the disk encryption set. I am pulling the PR into a WIP state until this is resolved. 
```
Details=[{"code":"BadRequest","message":"Required parameter for Confidential VMs 'managedDisk.securityProfile.securityEncryptionType' is missing (null)."}]
```

nywilken:
Hi @julian1059 thanks for opening up this change. Nice to see it closes two issues. I recently worked on enhancing the DES support for SIGs in #371 it is not merged yet but I suspect there will be conflicts. I'll take a look and see how things work. I would suggest giving the #371 a read through as well, as it may change how you refer to DES ids.

julian1059:
Hi @nywilken - No problem on the PR, happy to help. I just took a look at #371. I like what you did with the `target_region` blocks. I think my changes will integrate nicely into the new format with a couple of tweaks. 

From my POV, there are two things needed for Confidential VMs.
1. Deploying the packer build VM as security type ConfidentialVM
2. Publishing to the SIG with a ConfidentialVMEncryptionType set (matching the image def)

The first item appears to be unaffected from the `target_region` block change since the `disk_encryption_id` parameter (outside of the sig destination) handles any disk encryption needed for the build vm. 

The second item needs to be adjusted to use the disk encryption id in the target region struct. This will enhance the cvm logic since I can now safely include the encryption type of `EncryptedWithCmk` without assuming the same des for the build vm will be used for the current image version. That is awesome, I was not very thrilled with that assumption!

What is the timeline for getting #371 merged? After it is merged, I can adjust my code and handle any conflicts before a final review.

nywilken:
@julian1059 The PR has been merged and release please let me know if you have cycles to rebase and pick this work back up. Thanks again for the contribution.

julian1059:
@nywilken Great to hear! I have cycles to rebase and refactor this change a bit. I'll be taking a look over the next few days.

builder/azure/arm/template_factory.go
```diff
@@ -200,6 +201,29 @@ func GetVirtualMachineTemplateBuilder(config *Config) (*template.TemplateBuilder
...
+			}
+			fallthrough
+		default:
+			return nil, fmt.Errorf("template: invalid security type %s. Valid options are 'TrustedLaunch' or 'ConfidentialVM'", config.SecurityType)
+		}
+		err = builder.SetSecurityProfile(config.SecureBootEnabled, config.VTpmEnabled, config.EncryptionAtHost, chosenSecurityType)
+		if err != nil {
+			return nil, err
+		}
+	}
```
JenGoldstrich, line 225: The way we handle security type is a bit confusing here to me, as it's not clear to the user when a security type is required,  can we refactor this to not use break/fallthrough and make this error case a bit clearer

builder/azure/arm/template_factory.go
```diff
@@ -208,7 +232,14 @@ func GetVirtualMachineTemplateBuilder(config *Config) (*template.TemplateBuilder
...
+
+	if config.DiskEncryptionSetId == "" && chosenSecurityType != nil {
+		err = builder.SetDiskEncryptionWithPaaSKey(chosenSecurityType)
 		if err != nil {
 			return nil, err
 		}
```
JenGoldstrich, line 245: I think this would be a bit cleaner if we only called this function on confidential VM builds, rather than checking this conditional inside of the template builder.

```suggestion
	if config.DiskEncryptionSetId == "" && chosenSecurityType == hashiVMSDK.SecurityTypesConfidentialVM {
		err = builder.SetDiskEncryptionWithPaaSKey(chosenSecurityType)
		if err != nil {
			return nil, err
		}
```

JenGoldstrich (commented):
Hey @julian1059 I left some initial feedback after a first pass reading through this code, but I wanted to say thanks for taking this problem on! It's a tough one and something I know users want to use this plugin for, so I'm excited to work with you and get this functionality into the plugin!

Can you provide me some more guidance on how you are testing this?  I was unable to use this PR to build a ConfidentialVM SIG Image, once I enabled Confidential VMs in my SIG image I was unable to push, reading docs it appears that confidential VMs require a VHD source or a "gen 2 managed image" source, but when trying to use managed image I ran into errors, what are you using as your base image to build this confidential image?   

When trying to build a VHD with the image you referenced in the issue `Canonical/0001-com-ubuntu-confidential-vm-jammy/22_04-lts-cvm` I ran into errors saying VHDs were not supported for this image.

Generally I think we may want to adjust the structure of how we accept this field, and add more validation to this to make it clearer for users how to use confidential VMs, but I want to get this up and running so I can have a better understanding of what is required and what is not allowed by the Azure API

julian1059:
> Hey @julian1059 I left some initial feedback after a first pass reading through this code, but I wanted to say thanks for taking this problem on! It's a tough one and something I know users want to use this plugin for, so I'm excited to work with you and get this functionality into the plugin!
> 
> Can you provide me some more guidance on how you are testing this? I was unable to use this PR to build a ConfidentialVM SIG Image, once I enabled Confidential VMs in my SIG image I was unable to push, reading docs it appears that confidential VMs require a VHD source or a "gen 2 managed image" source, but when trying to use managed image I ran into errors, what are you using as your base image to build this confidential image?
> 
> When trying to build a VHD with the image you referenced in the issue `Canonical/0001-com-ubuntu-confidential-vm-jammy/22_04-lts-cvm` I ran into errors saying VHDs were not supported for this image.
> 
> Generally I think we may want to adjust the structure of how we accept this field, and add more validation to this to make it clearer for users how to use confidential VMs, but I want to get this up and running so I can have a better understanding of what is required and what is not allowed by the Azure API

@JenGoldstrich Thanks for taking a look at the PR, I am happy to help out. I found through this process that confidential VMs are quite nuanced, and support is limited on the Azure side. How I have been testing is simply allowing packer to use the VirtualMachine (confidential) as the source for the SIG image version, marked as confidential. I have been excluding output for a manged image or for a VHD going to a storage account. I have simply been listing the `shared_image_gallery_destination` block in my packer file.

```hcl
source "azure-arm" "linux_mktpl" {
  subscription_id        = var.subscription_id
  cloud_environment_name = var.cloud
  use_azure_cli_auth     = true

  # Marketplace image source data
  image_publisher = var.image_publisher
  image_offer     = var.image_offer
  image_sku       = var.image_sku


  # Shared image gallery destination data
    subscription   = var.subscription_id
    resource_group = var.resource_group_name
    gallery_name   = var.shared_image_gallery_name
    image_name     = local.image_name
    image_version  = local.date
    dynamic "target_region" {
      for_each = local.replication_regions
      content {
        name = target_region.value
      }
    }
    confidential_vm_image_encryption_type = local.sig_destination_confidential_vm_image_encryption_type
  }

  shared_image_gallery_replica_count               = local.shared_image_gallery_replica_count
  shared_gallery_image_version_exclude_from_latest = local.shared_image_gallery_exclude_from_latest
  shared_image_gallery_timeout                     = local.shared_image_gallery_timeout

  # Networking information
  virtual_network_name                = var.virtual_network_name
  virtual_network_subnet_name         = var.virtual_network_subnet_name
  virtual_network_resource_group_name = var.virtual_network_resource_group_name

  # Additional source data
  build_resource_group_name = var.resource_group_name
  os_type                   = var.os_type
  temp_compute_name         = local.temp_compute_name
  temp_os_disk_name         = "${local.temp_compute_name}_disk"
  temp_nic_name             = "${local.temp_compute_name}_nic"
  ssh_username              = local.username
  ssh_password              = var.SSH_PASSWORD != null ? base64decode(var.SSH_PASSWORD) : null
  vm_size                   = local.vm_size

  # Confidential VMs
  vtpm_enabled        = local.vtpm_enabled
  secure_boot_enabled = local.secure_boot_enabled
  security_type       = local.security_type
}
```
```console
==> azure-arm.linux_mktpl: Publishing to Shared Image Gallery ...
==> azure-arm.linux_mktpl:  -> Source ID used for SIG publish        : '/subscriptions/<GUID>/resourceGroups/rg-eus-image-builds/providers/Microsoft.Compute/virtualMachines/pk-u22-cvm'
==> azure-arm.linux_mktpl:  -> SIG publish resource group            : '<rg_name>'
==> azure-arm.linux_mktpl:  -> SIG gallery name                      : '<sig_name>'
==> azure-arm.linux_mktpl:  -> SIG image name                        : 'u22-cvm'
==> azure-arm.linux_mktpl:  -> SIG image version                     : '2024.02.28'
==> azure-arm.linux_mktpl:  -> SIG Confidential VM Encryption Type   : EncryptedVMGuestStateOnlyWithPmk
==> azure-arm.linux_mktpl:  -> SIG target regions                    : '[eastus]'
==> azure-arm.linux_mktpl:  -> SIG storage account type              : ''
==> azure-arm.linux_mktpl:  -> SIG image version endoflife date      : ''
==> azure-arm.linux_mktpl:  -> SIG image version exclude from latest : 'true'
==> azure-arm.linux_mktpl:  -> SIG replica count [1, 100]            : '1'
==> azure-arm.linux_mktpl:  -> Shared Gallery Image Version ID       : '/subscriptions/<GUID>/resourceGroups/<rg_name>/providers/Microsoft.Compute/galleries/<sig_name>/images/u22-cvm/versions/2024.02.20'
```

Moreover, deploying to a managed image and publishing to a shared image gallery is supported by the packer plugin, however, Azure cannot create a managed image from a CVM, therefore this error was included if the packer build vm has all of the properties set to create a CVM:
```console
* A managed image (managed_image_name, managed_image_resource_group_name) can not set SecureBoot or VTpm, these features are only supported when directly publishing to a Shared Image Gallery
```

To make things more confusing the SIG image def has two ways to support CVMs, enabled and supported. Enabled restricts the image def to CVM based sources, while supported allows the use of non CVM based sources for its image versions. Using the following SIG definitions the following tests were successful:
- CVM Enabled - SIG Image Def
  - CVM from marketplace as source, SIG version as destination (marked as encrypted vm guest state ONLY with pmk to match the definition)
- CVM Supported - SIG Image Def
  - VM (non-confidential) marketplace as source, SIG version as destination (cvm encryption disabled since it is not required)
  - Managed Image as source (from non CVM), SIG version as destination (cvm encryption disabled since it is not required)
  - VHD as source (from non CVM), SIG version as destination (cvm encryption disabled since it is not required)

builder/azure/arm/template_factory.go
julian1059: Yes I agree with this. I will refactor this code. Also I discovered this am that this is not being evaluated when trying to create a VHD (to a storage account) from a CVM source. The refactor will include support for this.

builder/azure/arm/template_factory.go
julian1059: Ditto 👍

JenGoldstrich:
Hey @julian1059 thanks for the detailed response, I'm starting to get the biggest picture of the feature, it is quite nuanced as you say!  I'm having trouble running the same kind of build actually, trying to capture from a VM into a SIG destination

I have a SIG with the `ConfidentialVmSupported` feature enabled as its Security Type, and with the following template

```
source "azure-arm" "windows-sig" {
  use_azure_cli_auth = true
  vtpm_enabled = true
  secure_boot_enabled = true
  security_type = "ConfidentialVM"
  image_name     = var.image_name
   gallery_name   = var.gallery_name
   image_version  = var.image_version
  resource_group = var.resource_group_name
   target_region {
    name = "East US"
   }
    confidential_vm_image_encryption_type = "EncryptedVMGuestStateOnlyWithPmk"
  }

  os_type         = "Linux"
  image_publisher = "Canonical"
  image_offer     = "0001-com-ubuntu-confidential-vm-jammy"
  image_sku       = "22_04-lts-cvm"
  location = "East US"
  vm_size  = "Standard_DC2as_v5"
}
```

Using your PR I get the following error when trying to publish to SIG

```
==> azure-arm.windows-sig:  -> SIG Confidential VM Encryption Type   : EncryptedVMGuestStateOnlyWithPmk
==> azure-arm.windows-sig:  -> SIG target regions                    : '[eastus]'
==> azure-arm.windows-sig:  -> SIG storage account type              : ''
==> azure-arm.windows-sig:  -> SIG image version endoflife date      : ''
==> azure-arm.windows-sig:  -> SIG image version exclude from latest : 'false'
==> azure-arm.windows-sig:  -> SIG replica count [1, 100]
==> azure-arm.windows-sig: performing CreateOrUpdate: galleryimageversions.GalleryImageVersionsClient#CreateOrUpdate: Failure sending request: StatusCode=0
 -- Original Error: Code="InvalidParameter" Message="The security type 'ConfidentialVmSupported' is not supported for the source id '/subscriptions/[hidden]/resourceGroups/[hidden]/providers/Microsoft.Compute/virtualMachines/pkrvmeyyc50ozf5'. 
 Currently only Vhd Blob, User Image and Gallery Image Version sources are supported for 'ConfidentialVmSupported' images." Target="galleryImageVersion.properties.storageProfile.osDiskImage.source.id"
```

This is the description of my SIG image definition inside of the Azure API
```
[
  {
    "architecture": "x64",
    "features": [
      {
        "name": "SecurityType",
        "value": "ConfidentialVmSupported"
      }
    ],
    "hyperVGeneration": "V2",
    "id": "/subscriptions/[redacted]/resourceGroups/jenna/providers/Microsoft.Compute/galleries/jenna2/images/jenna",
    "identifier": {
      "offer": "0001-com-ubuntu-confidential-vm-jammy",
      "publisher": "Canonical",
      "sku": "22_04-lts-cvm"
    },
    "location": "eastus",
    "name": "jenna",
    "osState": "Generalized",
    "osType": "Linux",
    "provisioningState": "Succeeded",
    "resourceGroup": "jenna",
    "tags": {},
    "type": "Microsoft.Compute/galleries/images"
  }
]
```

What am I missing 🤔? I think it may either be the wrong VM size or a wrong OS but the error is very misleading

julian1059:
@JenGoldstrich - I believe the issue is your SIG image definition. Its security type is set to `ConfidentialVmSupported` and not `ConfidentialVM`. I see a similar error using the Azure portal, Azure will not allow a CVM as input to a CVM-supported definition; I will attach a screenshot. As I noted above, I have had success pushing image versions to CVM-supported image defs that are sourced from non-confidential VMs. 

Screenshot 2024-02-29 at 10 17 03 AM

JenGoldstrich:
Gotcha gotcha, I was able to build a confidential VM image version successfully, thanks for clarifying that, I see now from your list of successful builds, I'll run some more builds and leave some more feedback now that I understand that, thanks for clarifying that!

builder/azure/common/template/template_builder.go
```diff
@@ -316,15 +317,39 @@ func (s *TemplateBuilder) SetOSDiskSizeGB(diskSizeGB int32) error {
...
 	if err != nil {
 		return err
 	}
 
 	profile := resource.Properties.StorageProfile
-	profile.OsDisk.ManagedDisk.DiskEncryptionSet = &DiskEncryptionSetParameters{
-		ID: &diskEncryptionSetID,
+	if securityType != nil && *securityType == hashiVMSDK.SecurityTypesConfidentialVM {
+		profile.OsDisk.ManagedDisk.SecurityProfile = &virtualmachines.VMDiskSecurityProfile{}
+		securityEncryptionType := virtualmachines.SecurityEncryptionTypesDiskWithVMGuestState
```
JenGoldstrich, line 329: Here and at line 351 you're setting the VM's security ecryption type to DiskWithVMGuestState, whenever I try to build with a CMK while setting a DES, I run into the following error 

```
==> azure-arm.windows-sig: polling after CreateOrUpdate: Code="Conflict" Message="The SecurityProfile type of this image version 'EncryptedWithCmk' must match the SecurityProfile type of the source 'EncryptedVMGuestStateOnlyWithPmk'."
```

When I try to build with EncryptedWithPmk instead I get a similar error
```
==> azure-arm.windows-sig: polling after CreateOrUpdate: Code="Conflict" Message="The SecurityProfile type of this image version 'EncryptedWithPmk' must match the SecurityProfile type of the source 'EncryptedVMGuestStateOnlyWithPmk'."
```

Because the source VM will always be using guest state only encryption type, I think we should set this based on the value supplied in the confidential_vm_image_encryption_type field, and only assume Guest State if its unset, or maybe better yet require the user to explicitly select an encryption type, what do you think?

builder/azure/arm/template_factory.go
julian1059: Refactored in [617b8d3](https://github.com/hashicorp/packer-plugin-azure/pull/373/commits/617b8d33cb53b8e75e3146043d4c3fa536cc60c4) ✅

builder/azure/arm/template_factory.go
julian1059: @JenGoldstrich After digging back into the code I remembered why I was checking the chosenSecurityType in the builder rather than the factory. `hashiVMSDK.SecurityTypesConfidentialVM` is a const making an inline pointer deference impossible. In my latest commit, I removed this clause from the conditional and instead chose to ensure the pointer was not nil in the builder.

Will this work? I recognize you asked me to move logic out of the template builder and I did the exact opposite. I can adjust this again if you think it's still necessary.

builder/azure/common/template/template_builder.go
julian1059: @JenGoldstrich Thanks for bringing this up, at my first go at this I assumed that `VMGuestStateOnly` and `DiskWithVMGuestState` were mutually exclusive to either PaaS key encryption or CMK key encryption. After checking these errors out and doing some research, I agree that users should be able to supply this setting. Then this setting will be used for either PaaS or CMK-based encryption. I prefer this option to pulling from the `confidential_vm_image_encryption_type` field since that field is tightly coupled to the SIG image version.

I will get this implemented and tested tomorrow.

JenGoldstrich: Hey @julian1059, taking another look now

That's fair, my concern is that now `security_encryption_type` is tightly coupled with confidential_vm_image_encryption type, there are a lot of cases where builds can fail now that we'd have to add extra validation cases for, or document around, but if we infer one of these fields from the other then users don't have to set as many fields to build a confidential VM, and there's less errant states for the template to be in.  It feels like this feature requires a lot of fields to be set, and users have to figure out that EncryptedWithPMK expects `DiskWithVMGuestState`.

One way to simplify this in my mind would be as a SIG Destination option to have a field if the image verison is encrypted with a PMK or CMK, and then infer the VM Guest State from the security_encryption_type, since it must always match the encryption state of the image version.  Maybe a field like `use_platform_key_encryption`, what do you think?

JenGoldstrich: Also related but I was able to build both VMGuestStateOnly and DiskWithVMGuestState images via PMKs today with this build, nice work!

builder/azure/arm/template_factory.go
JenGoldstrich: Yeah totally, this is still cleaner since we're handling those checks in one place, no worries and thanks for explaining!

builder/azure/common/template/template_builder.go
JenGoldstrich: Actually after thinking about this more we should probably keep these separated and just add more validation around this to ensure that there isn't a mismatch between these fields, especially if Azure ever adds other encryption types which they can always do this keeps us more flexible 🤔 , I'll leave my original message just to keep the thought process but I think instead we could just ensure that there isn't a mismatch of the SIG encryption type and the security_encryption_type in the config validation.

julian1059: It's awesome to hear that you successfully built both VMGuestStateOnly and DiskWithVMGuestState images via PMKs! 👌 

Regarding security_type & security_encryption_type, thank you for outlining your thought process here. I agree with your last note, we should leave these two fields separate. The original goal was to simplify as much as possible, but the simplification seems to require assumptions, so yes keeping these fields separate seems ideal at this point.

Regarding validations, are there additional validations you have found other than the ones I listed below? Also, what happens if the inputs are unknown constants to the SDK (for future proofing scenarios), for example, say, `EncryptedVMGuestStateOnlyWithCmk`?
```go
// DiskWithVMGuestState needs to match EncryptedWithPMK
// DiskWithVMGuestState needs to match EncryptedWithCMK
// VMGuestStateOnly needs to match EncryptedVMGuestStateOnlyWithPmk
```

JenGoldstrich: Your suggested validations sound good to me, I would add validations to check that the three new fields (Security type, security encryption type, and Confidential VM image encryption type) are only the expected values we support from the SDK (i.e. only. EncryptedWithPMK, EncryptedWithCMK, and EncryptedVMGuestStateOnlyWithPmk are allowed for confidential vm image encryption type)

One more validation that might be worth it that I'm less confident in and would like to hear your opinion is validation to only allow `confidential_vm_image_encryption_type` to be set when security type is confidential, since this field is ignored otherwise.

In terms of future proofing if we have a clean error in the validator that says for example for security_encryption_type "Only DiskWithVMGuestState or VMGuestStateOnly is supported", it'll make it easy in the future to quickly identify why this validation was added and where in the code we'd need to change to support a new field, I wouldn't worry about future proofing unknown constants past that, if we combined the variables I think this would be less future proof.

julian1059:
@JenGoldstrich Let me know what you think about the validations. Here are three test cases I tried and the results I saw after the validation changes were added. 

### Test 1
```hcl
security_type            = "SecurityTypeNotAllowed"
security_encryption_type = "VMGuestStateOnly"

  confidential_vm_image_encryption_type = "EncryptedWithCmk"
}
```
```console
❯ packer build -only=azure-arm.linux_mktpl . 
Error: 3 error(s) occurred:

* The security_type "SecurityTypeNotAllowed" must match either "TrustedLaunch" or "ConfidentialVM".
* security_encryption_type must be unset if the security_type is not set to "ConfidentialVM".
* The security_encryption_type setting "VMGuestStateOnly" does not match the shared_image_gallery_destination setting "EncryptedWithCmk". security_encryption type "VMGuestStateOnly" needs to match "EncryptedVMGuestStateOnlyWithPmk".

  on azure_source.pkr.hcl line 50:
  (source code not available)
```

### Test 2
```hcl
security_type            = "ConfidentialVM"
security_encryption_type = "DiskWithVMGuestState"

  confidential_vm_image_encryption_type = "EncryptedVMGuestStateOnlyWithPmk"
}
```
```console
❯ packer build -only=azure-arm.linux_mktpl . 
Error: 1 error(s) occurred:

* The security_encryption_type setting "DiskWithVMGuestState" does not match the shared_image_gallery_destination setting "EncryptedVMGuestStateOnlyWithPmk". security_encryption type "DiskWithVMGuestState" needs to match "EncryptedWithPMK" or "EncryptedWithCMK".

  on azure_source.pkr.hcl line 50:
  (source code not available)
```

### Test 3
```hcl
security_type            = "TrustedLaunch"
security_encryption_type = "DiskWithVMGuestState"

  confidential_vm_image_encryption_type = "EncryptedWithPmk"
}
```
```console
❯ packer build -only=azure-arm.linux_mktpl . 
Error: 1 error(s) occurred:

* security_encryption_type must be unset if the security_type is not set to "ConfidentialVM".

  on azure_source.pkr.hcl line 50:
  (source code not available)
```

julian1059:
@JenGoldstrich Our current validations include:

- Validate that ConfidentialVM sources are not used to build managed images or VHDs
- Validate SecurityType matches either TrustedLaunch or ConfidentialVM
- Validate SecurityEncryptionType matches one of the values in the sdk
- Validate SigDestinationConfidentialVMImageEncryptionType matches one of the values in the sdk
- Validate that the SecurityEncryptionType is not set if the SecurityType is not set to ConfidentialVM
- Validate DiskWithVMGuestState matches EncryptedWithPMK or EncryptedWithCMK
- Validate VMGuestStateOnly matches EncryptedVMGuestStateOnlyWithPmk

I thought of two additional validations overnight and took the liberty of adding them in [37e9fc2](https://github.com/hashicorp/packer-plugin-azure/pull/373/commits/37e9fc24911b5c6ce504564191d6bbb0276118c4).

- SharedGalleryDestination: Ensure no disk encryption set is set in the target regions when not using CMK encryption
```console
❯ packer build -only=azure-arm.linux_mktpl . 
Error: 1 error(s) occurred:

* confidential_vm_image_encryption_type must be set to "EncryptedWithCmk" when passing a disk_encryption_set_id in the target_region block

  on azure_source.pkr.hcl line 50:
  (source code not available)
```
- Ensure if the sig destination encryption type is set to EncryptedWithCmk, and the source vm encryption type is set to ConfidentialVM, that a disk encryption set is provided for the source vm
```console
❯ packer build -only=azure-arm.linux_mktpl . 
Error: 1 error(s) occurred:

* when using a confidential vm as source to an cvm image version and a confidential_vm_image_encryption_type of "EncryptedWithCmk", the source cvm must have a disk_encryption_set_id set

  on azure_source.pkr.hcl line 50:
  (source code not available)
```

Regarding your question about only allowing confidential_vm_image_encryption_type to be set when the security type is confidential. I am hesitant to include that validation in case users want to publish a VHD to a storage account and then use that VHD as a source for a CVM image. Azure notes that CVM sources can be snapshots, VHDs, and VM sources. I will test this once more and report back findings.

builder/azure/arm/config.go
```diff
@@ -1291,6 +1309,18 @@ func assertRequiredParametersSet(c *Config, errs *packersdk.MultiError) {
...
 	}
 
+	if c.SecurityType != "" {
+		if ok, err := assertAllowedSecurityType(c.SecurityType, "security_type"); !ok {
+			errs = packersdk.MultiErrorAppend(errs, err)
+		}
+	}
+
+	if c.SecurityEncryptionType != "" {
+		if ok, err := assertAllowedSecurityEncryptionType(c.SecurityEncryptionType, "security_encryption_type"); !ok {
```
JenGoldstrich, line 1319: I like how you've broken out this and `assertAllowedSecurityType` into their own functions, but I think the second parameter in both of these functions might not be needed, since they are only called once with those parameters, I think you can just hardcode that field name in the string

builder/azure/arm/config.go
```diff
@@ -1324,6 +1354,26 @@ func assertRequiredParametersSet(c *Config, errs *packersdk.MultiError) {
...
+			for _, r := range c.SharedGalleryDestination.SigDestinationTargetRegions {
+				if r.DiskEncryptionSetId != "" && c.SharedGalleryDestination.SigDestinationConfidentialVMImageEncryptionType != string(galleryimageversions.ConfidentialVMEncryptionTypeEncryptedWithCmk) {
+					errs = packersdk.MultiErrorAppend(errs, fmt.Errorf("confidential_vm_image_encryption_type must be set to \"EncryptedWithCmk\" when passing a disk_encryption_set_id in the target_region block"))
+				}
+			}
+
+			// Ensure if the encryption type is set to EncryptedWithCmk, and the encryption type is set to ConfidentialVM, that the a disk encryption id is set
+			if c.SecurityType == constants.ConfidentialVM && c.SharedGalleryDestination.SigDestinationConfidentialVMImageEncryptionType == string(galleryimageversions.ConfidentialVMEncryptionTypeEncryptedWithCmk) {
+				if c.DiskEncryptionSetId == "" {
+					errs = packersdk.MultiErrorAppend(errs, fmt.Errorf("when using a confidential vm as source to an cvm image version and a confidential_vm_image_encryption_type of \"EncryptedWithCmk\", the source cvm must have a disk_encryption_set_id set"))
```
JenGoldstrich, line 1373: Nice! This is an awesome case to add and will save users a lot of time, good thinking!

JenGoldstrich (commented):
Validations are looking great! I think we should add unit tests to them to `config_test.go` for the `Should Reject` cases, let me know the results of your VHD source research and if you'd like any help updating those tests, I am planning to run acceptance tests against your PR next week!

builder/azure/arm/config.go
julian1059: Yep, great point. I will remove the param in my next commit.

julian1059:
@JenGoldstrich Great to hear your positive feedback, I am excited to get this PR implemented. Ill add the failure validation tests to the `config_test.go` file at some point this afternoon.

builder/azure/arm/config.go
julian1059: Okay, those parameters were removed, and the setting value was instead hard-coded in the error string — changes made in [f26bf29](https://github.com/hashicorp/packer-plugin-azure/pull/373/commits/f26bf290ba52b1bb16aa438428f9ab26edea3324).

builder/azure/arm/config.go
julian1059: I also was able to add an else statement here, to validate the disk_encryption_ids in the target region blocks!

julian1059:
@JenGoldstrich The test cases have been added! I added another validation to ensure if:

1. the sig destination encryption type is set to EncryptedWithCmk
2. the source VM encryption type is set to ConfidentialVM
3. a disk encryption set is provided for the source VM

that a disk encryption set must also be provided for each of the target region blocks. Very specific case but thought it may save some headaches.

I also finished my research on the VHD as a source for a SIG image. The plugin does not currently support a VHD as a source for a SIG image. I think we _could_ add validation to ensure a SIG destination cvm encryption type is only set when the security type is set to cvm. However, I don't know how much it would benefit us. I am taking the position that the source VM and SIG destination are mutually exclusive and that the onus should be on the user to use the correct security settings for both. What do you think? The worst-case scenario is that a user will get an ARM error describing a VM source and image def destination security type mismatch. This also ensures the two feature additions are distinct.

builder/azure/arm/config_test.go
```diff
@@ -3024,3 +3024,290 @@ func TestConfigShouldRejectSharedImageGalleryDestinationReplicationRegions(t *te
...
+		"os_type":                           constants.Target_Linux, // Does not matter for this test case, just pick one.
+		"security_type":                     "ConfidentialVM",
+		"security_encryption_type":          "VMGuestStateOnly",
+	}
+
+	var c Config
+	_, err := c.Prepare(config, getPackerConfiguration())
+	if err == nil {
+		t.Fatalf("expected config to reject with the following error: %q", "Setting a security type of \"ConfidentialVM\" is not allowed when building a VHD or creating a Managed Image, only when publishing directly to Shared Image Gallery")
+	}
```
JenGoldstrich, line 3047: Can you add to these tests checks on the specific error message like the other tests in this file so this would become

```suggestion
	_, err := c.Prepare(config, getPackerConfiguration())
	errMessage := `Setting a security type of \"ConfidentialVM\" is not allowed when building a VHD or creating a Managed Image, only when publishing directly to Shared Image Gallery`
	if err == nil {
		t.Fatalf("expected config to reject with the following error: %s", errMessage)
	}
	if !strings.Contains(err.Error(), errorMessage) {
		t.Errorf("expected config to reject with error containing %s but got %s", errorMessage, err)
	}
```

Happy to help make this change too

JenGoldstrich (commented):
Left a comment on the test coverage, today I tested several cases for PMK and ran the acceptance tests which passed 🥳, tommorow I will follow up on the unit test suggestion I left, as well as test CMK cases, if all is well though I'm hoping to get this approved soon, thank you for your patience and willingness to iterate through a few rounds of feedback on this one, I think it'll make the experience of using this feature a lot better because of your work.

builder/azure/arm/config_test.go
julian1059: Ah ya this is much cleaner. Ill take you up on the help here! Feel free to commit to my branch. Appreciate the assist. 👍

JenGoldstrich (approved):
When trying to use multiple target regions Confidential VM builds fail with the following error

```
==> azure-arm.windows-sig: performing CreateOrUpdate: galleryimageversions.GalleryImageVersionsClient#CreateOrUpdate: Failure sending request: StatusCode=0 -- Original Error: Code="InvalidParameter" Message="Only the home region 'eastus' can be specified as a target region for Images with the ConfidentialVM feature unless the Confidential VM encryption type is set to
'NonPersistedTPM'." Target="galleryImageVersion.properties.publishingProfile.targetRegions"
```

`NonPersistedTPM` was introduced as an option after the version of the Gallery Image Versions API then is currently in use by the plugin, this will require a bit more lift.

I don't want to delay merging this since everything else is working, and it'd be good to get feedback on this feature while working on upgrading the version of the SDK we use, I'm approving this now and will address this in a follow up PR.

Thanks a lot for your work and willingness to iterate through several rounds of feedback, I am excited to finally get this change merged!  Great job and thanks for the contribution!

PR #381, opened 2024-02-22
Add default build location when no target region is set

nywilken:
Previously users were able to build to a Shared Image Gallery without
specifying a primary target region, as it would default to the build
location. This functionality was removed for the `target_region` block
attribute in hopes that users would be more explicit on which regions to
publish to. However, not having a primary target region breaks the
existing behavior for publishing with no region set.

To avoid further breaking changes support for automatically adding a
default target region is being added back but with a WARNING indicating
that the behavior will change in a future release.

### DIFF
--- a/builder/azure/arm/builder.go
+++ b/builder/azure/arm/builder.go
@@ -248,7 +248,9 @@ func (b *Builder) Run(ctx context.Context, ui packersdk.Ui, hook packersdk.Hook)
 		}
 
 		if len(b.config.SharedGalleryDestination.SigDestinationTargetRegions) == 0 {
-			return nil, errors.New("no target destination region specified for the Shared Image Gallery; use target_region to specify at least the primary destination region for storing the image version.")
+			buildLocation := normalizeAzureRegion(b.stateBag.Get(constants.ArmLocation).(string))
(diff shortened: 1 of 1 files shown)


builder/azure/arm/builder.go
```diff
@@ -248,7 +248,9 @@ func (b *Builder) Run(ctx context.Context, ui packersdk.Ui, hook packersdk.Hook)
 		}
 
 		if len(b.config.SharedGalleryDestination.SigDestinationTargetRegions) == 0 {
-			return nil, errors.New("no target destination region specified for the Shared Image Gallery; use target_region to specify at least the primary destination region for storing the image version.")
+			ui.Say("WARNING! No target region was set for the Shared Image Gallery. Adding %q as the primary target region for now but this may change in the future.")
```
lbajolet-hashicorp, line 251: You forgot to add a `Sprintf` here to expand the `%q` I believe?

nywilken: No I didn't 😉 - good catch.

lbajolet-hashicorp (approved):
LGTM! This is still a draft though, are there more changes coming up on this?

PR #383, opened 2024-02-24
Update validation check for Shallow Replication

nywilken:
Shallow replication for a SIG allows for there to be no target region
set. But if there are more than one region defined it should fail, as
only one region matching the build region can be set when using shallow
replication.


JenGoldstrich (approved):
LGTM after our pairing!

Issue #386, opened 2024-03-08
Add ability to set replica count per region for each shared image gallery destination region


nywilken:
#### Community Note

Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request.
Please do not leave "+1" or "me too" comments, they generate extra noise for issue followers and do not help prioritize the request.
If you are interested in working on this issue or have submitted a pull request, please leave a comment.

#### Description

When publishing to a Shared Image Gallery using the `target_region` block I would like 
to have an option for setting individual replica counts per regions as opposed to on global count in
`shared_image_gallery_replica_count`. 

#### Use Case(s)

Users may want to set different replica counts per region to reduce cost and resiliency per region. 

#### Potential configuration

```hcl
  shared_image_gallery_destination {
    subscription         = var.subscription_id
    resource_group       = var.gallery_resource_group
    gallery_name         = var.gallery
    image_name           = var.box_name
    image_version        = var.version
    storage_account_type = "Standard_LRS"
    target_region   {
       name = "us-east-1"
       replicas = 20
    }
    target_region   {
       name = "us-west-2"
       replicas = 5
    }
  }
```

#### Potential References

https://discuss.hashicorp.com/t/packer-arm-regionalreplicacount/63513

PR #387, opened 2024-03-08
Add ability to set individual replica count per region

nywilken:
This changes allows a user to set different replica counts for each
target region defined with a shared_image_gallery_destination block.

Closes: https://github.com/hashicorp/packer-plugin-azure/issues/386

### DIFF
--- a/builder/azure/arm/config.go
+++ b/builder/azure/arm/config.go
@@ -146,6 +146,8 @@ type TargetRegion struct {
 	// the replication of encrypted disks across regions. CMKs must
 	// already exist within the target regions.
 	DiskEncryptionSetId string `mapstructure:"disk_encryption_set_id"`
+	// Number of replicas in this region. Default: 1
+	ReplicaCount int64 `mapstructure:"replicas"`
 }
 
 type Spot struct {
--- a/builder/azure/arm/config.go
+++ b/builder/azure/arm/config.go
@@ -146,7 +146,9 @@ type TargetRegion struct {
 	// the replication of encrypted disks across regions. CMKs must
 	// already exist within the target regions.
 	DiskEncryptionSetId string `mapstructure:"disk_encryption_set_id"`
-	// Number of replicas in this region. Default: 1
+	// The number of replicas of the Image Version to be created within the region. Defaults to 1.
+	// Replica count must be between 1 and 100, but 50 replicas should be sufficient for most use cases.
+	// When using shallow replication `use_shallow_replication=true` the value can only be 1 for the primary build region.
 	ReplicaCount int64 `mapstructure:"replicas"`
 }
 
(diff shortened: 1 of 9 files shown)


nywilken:
It looks like the Azure SDK allows for there to be a global replica count as well as a regional replica count. I need to verify what happens when both values are set.

builder/azure/arm/config.go
```diff
@@ -1334,22 +1337,21 @@ func assertRequiredParametersSet(c *Config, errs *packersdk.MultiError) {
 		}
 		// Validate target region settings; it can be the deprecated replicated_regions attribute or multiple target_region blocks
 		if (len(c.SharedGalleryDestination.SigDestinationReplicationRegions) > 0) && (len(c.SharedGalleryDestination.SigDestinationTargetRegions) > 0) {
-			errs = packersdk.MultiErrorAppend(errs, errors.New("`replicated_regions` can not be defined alongside `target_region`; you can defined a target_region for each destination region you wish to replicate to."))
+			errs = packersdk.MultiErrorAppend(errs, errors.New("`replicated_regions` can not be defined alongside `target_region`; you can define a target_region for each destination region you wish to replicate to."))
+		}
+
+		if (len(c.SharedGalleryDestination.SigDestinationTargetRegions) > 0) && c.SharedGalleryImageVersionReplicaCount != 0 {
+			errs = packersdk.MultiErrorAppend(errs, errors.New("`shared_image_gallery_replica_count` can not be defined alongside `target_region`; you can specify the number of replicas per region within target_region blocks."))
```
JenGoldstrich, line 1344: Two notes on this I found the error message a bit confusing when talking about `single target_region blocks` so I reworded that

worth noting this is a breaking change since users would have had to set the replica count with shared_image_gallery_replica_count before, even when using a target_region block, so its worth making the next release a minor version bump, and calling this out in the release notes I think

JenGoldstrich (approved):
I rebased this and there was one missing test case and an error message I felt was a bit confusingly worded so I changed that, otherwise this LGTM,

dhananjaipai:
getting the error
```
An argument named "ReplicaCount" is not expected here
```
when trying to use
```
dynamic "target_region" {
      for_each = split(",", var.target_regions)
      content {
        name                 = target_region.value
        ReplicaCount         = 1
      }
    }
```

I tried different combinations of regional_replica_count, regionalReplicaCount, ReplicaCount and others, can you improve the documentation so that the target_region spec is more clear?

currently is shows only
```
target_region ([]TargetRegion) - A target region to store the image version in.
```
you then expect to see something like [this](https://learn.microsoft.com/en-us/dotnet/api/azure.resourcemanager.compute.models.targetregion?view=azure-dotnet#properties) but nothing works?

dhananjaipai:
Reading through the commits found [this](https://github.com/hashicorp/packer-plugin-azure/pull/387/commits/90cd05f2a0123f541a801e18d761eb7344745762)

apparently the key was modified to be `replicas` ? 
and I can see zero references to that in the [docs](https://developer.hashicorp.com/packer/integrations/hashicorp/azure/latest/components/builder/arm) 

I can create a separate issue on this.

PR #391, opened 2024-03-14
Fixed typos

lzandman:
Fixed some typos I noticed.

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -107,7 +107,7 @@ output:
 # all available settings of specific linters
 linters-settings:
  errcheck:
-    # report about not checking of errors in type assetions: `a := b.(MyStruct)`;
+    # report about not checking of errors in type assertions: `a := b.(MyStruct)`;
     # default is false: such cases aren't reported by default.
     check-type-assertions: false
 
--- a/.web-docs/README.md
+++ b/.web-docs/README.md
@@ -47,7 +47,7 @@ Config allows for various ways to authenticate Azure clients.  When
 one of the following: `client_secret`, `client_jwt`, `client_cert_path` --
 Packer will use the specified Azure Active Directory (AAD) Service Principal
 (SP).
-If none ofthese options are specified, Packer will attempt to use the Managed Identity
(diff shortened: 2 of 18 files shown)


lzandman:
Why does that validation step fail? Did I alter some files that I shouldn't have?

lbajolet-hashicorp:
Hi @lzandman,

The files in `.web-docs` are generated from a combination of text in the `documentation` directory, and partials extracted from the Go code documentation (`packer-sdc` is the component responsible for that).

So in this case yes, those changes you made to the files in `.web-docs` are overwritten by `make generate`, which we run in CI so we always have up-to-date documentation.

I suggest you run the command locally and search for the source of those typos in the code, if fixed here, the generation should stabilise.
Let me know if you need help!

Thanks for the patch.

lzandman:
I'm not able to do a `make generate` on the system I currently work on. I did revert all files in the `.web-docs` folder.

lbajolet-hashicorp (approved):
Hi @lzandman,

Thanks for the update; I've taken the liberty to run `make generate` to update the `.web-docs` from your changes, hopefully the CI should be green.

If it succeeds, I'll merge the PR today.

Thanks again!

lzandman:
Really weird how this PR's diff view still shows changed files in `.web-docs`. I'm pretty sure my branch doesn't. Is the build that's currently running (actions) generating those files and adding them to this PR?

lbajolet-hashicorp:
Oh maybe I wasn't explicit when I commented sorry about that.

Everything in `.web-docs` is versioned, the documentation for the plugin is extracted from this directory.
We have a check in our pipelines for making sure the `.web-docs` generated data is up-to-date compared to the code (since comments are extracted as partials), and if there's a discrepancy between what was committed and what is generated in CI, this means that the `.web-docs` should be generated and committed to git.

Hope that clears-up that misunderstanding!

lzandman:
Ah, I understand.

lbajolet-hashicorp (approved):
LGTM\!

Issue #395, opened 2024-03-25
Add Support for `NonPersistantTMP` Option 


JenGoldstrich:
Follow up to https://github.com/hashicorp/packer-plugin-azure/pull/373

Confidential VMs can't be published to multiple regions currently due to the absence of `NonPersistantTPM` option in current galleryimageversion API

The following needs to be done to solve this issue
1.  Submit a PR to [Pandora](https://github.com/hashicorp/pandora) Azure SDK generation tool to add 2024-03-01 API version which contains the above option.
2.) Update code base to use new version of the SDK, this will require updating a few different clients.
3.) Add logic to parse NonPersistantTPM option, and set it in Confidential VM Encryption type setting, ensuring that it allows multi-region deploys

JenGoldstrich:
that was the wrong API version actually oops, https://github.com/hashicorp/pandora/pull/4032 new PR for Pandora here

JenGoldstrich:
So this feature is actually not in a stable API yet in the virtual machines API, so we have decided to hold off until it is published in a stable API and then we will add support for this feature.

PR #396, opened 2024-03-26
Update to the newest version of hashicorp/go-azure-sdk

JenGoldstrich:
This PR makes several changes to upgrade `github.com/hashicorp/go-azure-sdk` from version  `v0.20230523.1140858` to the newest version at the time of this PR `v0.20240327.1161949`, and to fully remove the deprecated Microsoft `go-autorest` library.

1. Every request on the new SDK client layer requires a context with a deadline, otherwise requests fail with the following error `the context used must have a deadline attached for polling purposes, but got no deadline`, even if counter-intuitively the request would not use polling

2. When the tenant ID is unset and not using CLI auth the plugin makes an unauthenticated HTTP request to the GetSubscriptions endpoint to fetch the tenant ID through the auth headers, the new SDK does not return these headers unless the auth object is correctly configured so now this is done with a regular http client

3. The Chroot Builder used an autorest client to connect to the [Azure IMDS](https://learn.microsoft.com/en-us/azure/virtual-machines/instance-metadata-service?tabs=windows), this has also been modified to use a regular http client, enabling the removal of auto rest

4. The chroot builder has a long running bug https://github.com/hashicorp/packer-plugin-azure/issues/87 where we get an error that the `vm` template function is not defined unless setting 4 temporary values, this is due to an issue with the packer SDK wiping out this function when calling Decode, @lbajolet-hashicorp is investigating this issue further on the SDK end but to fix this on chroot this PR moves the function after Decode, so that is available in the builder.  The `vm` function does not appear to be working at all currently, but this seems due to the SDK bug, and will be addressed in a follow up PR if not.

Closes https://github.com/hashicorp/packer-plugin-azure/issues/87
Unblocks https://github.com/hashicorp/packer-plugin-azure/issues/395


builder/azure/common/client/azure_client_set.go
```diff
@@ -50,12 +45,18 @@ type AzureClientSet interface {
 var _ AzureClientSet = &azureClientSet{}
 
 type azureClientSet struct {
```
JenGoldstrich, line 47: While this client is in the common package it is only used by the chroot builder, the way I made this work with the new clients is a bit hacky, but I felt it wasn't worth a lot of time refactoring the chroot builder heavily as it makes up a very small percent of our usage.

go.mod
```diff
@@ -1,40 +1,36 @@
 module github.com/hashicorp/packer-plugin-azure
 
-go 1.19
+go 1.21
```
JenGoldstrich, line 3: https://github.com/hashicorp/packer-plugin-azure/blob/36cc195a84e36f246666d375c1e334fd8a5ef5ed/.go-version#L1

nywilken: Is 1.21.8 a requirement for these changes or should we just always set to 1.21.N since this is technically the latest version?

nywilken (approved):
This looks good to me. I appreciate the added context for the context.WithTimeout.

go.mod
JenGoldstrich: Not at all, I just bumped us to 1.21 based on what was already set in the .go-version, these changes should work independent of go minor version.

PR #398, opened 2024-04-01
Use a prefix to avoid resource unique name failures in acceptance tests

JenGoldstrich:
Recently acceptable tests in CI have been failing when I left behind some resources in the Packer team's shared test subscription, these resources, such as DTLs or Compute Galleries, have a unique name constraint on a subscription level.  The way these tests were setup in https://github.com/hashicorp/packer-plugin-azure/pull/346 was done because of how Packer and Terraform handle environment variables. This leads to less confidence in actual CI failures caused by regressions.

To handle this I've introduced the `ARM_RESOURCE_PREFIX` environment variable, and added a script to pass variables into Terraform more cleanly, this script can be run in CI to make sure the required variables are set, and then run terraform apply with the environment variables as command line arguments.

Successful example run on this branch: https://github.com/hashicorp/packer-plugin-azure/actions/runs/8529791780/job/23367911714


builder/azure/dtl/builder_acc_test.go
```diff
@@ -62,93 +65,8 @@ func TestDTLBuilderAcc_ManagedDisk_Linux_Artifacts(t *testing.T) {
 	})
 }
 
-const testBuilderAccManagedDiskWindows = `
```
JenGoldstrich, line 65: I had to change these templates to add the prefix variable, so I figured I'd just go ahead and extract these into template files

terraform/variables.tf
```diff
@@ -8,15 +8,14 @@ variable "resource_group_location" {
...
 }
 
 variable "storage_account_name" {
   type    = string
-  default = "packeracctest"
 }
 
-variable "dtl_name" {
+// Variable applied to resources that have uniqueness constraints at a subscription level
+// For example you can't have to shared image galleries named `linux` in the same Subscription in different resource group
```
sylviamoss, line 18:
```suggestion
// For example you can't have two shared image galleries named `linux` in the same Subscription in different resource group
```

sylviamoss (approved):
LGTM 👍🏼

PR #407, opened 2024-04-16
Fix panic when encrypted images or using security type with a Shared Image Gallery source

### LINKED ISSUE #406: Confidential VM builds - Packer azure plugin crash unexpected EOF
Hi,

We are trying to build an image on Azure using packer.

Our configuration file looks like this :

```
variable "prefix" {}
variable "client_id" {}
variable "tenant_id" {}
variable "subscription_id" {}
variable "client_secret" {}

locals {
  client_id = "${var.client_id}"
  client_secret = "${var.client_secret}"
  subscription_id = "${var.subscription_id}"
  tenant_id = "${var.tenant_id}"
  ubuntu_managed_image_name = "${var.prefix}-***"
  ubuntu_build_resource_group_name = "***"
  os_type = "***"
  image_publisher = "***"
  image_offer = "***"
  image_sku = "***"
  vm_size = "***"
  vtpm_enabled = true
  secure_boot_enabled = true
  security_type = "***"
}

source "azure-arm" "ubuntu" {
  client_id                 = local.client_id
  tenant_id                 = local.tenant_id
  subscription_id           = local.subscription_id
  client_secret             = local.client_secret
  build_resource_group_name   = local.ubuntu_build_resource_group_name
  os_type                     = local.os_type
  vm_size                     = local.vm_size
  secure_boot_enabled         = local.secure_boot_enabled
  vtpm_enabled                = local.vtpm_enabled
  security_type               = local.security_type
  shared_image_gallery {
    subscription = "***"
    resource_group = "***"
    gallery_name = "***"
    image_name = "***"
    image_version = "***"
  }
    resource_group = "***"
    gallery_name = "***"
    image_name = "***"
    image_version = "***"
    storage_account_type = "***"
    target_region {
      name = "***"
    }
  }
}

build {
  sources = ["source.azure-arm.ubuntu"]

}

```
When running `packer build -debug azure-sev-ubuntu.pkr.hcl`, Packer crashes even before trying to build the image, it seems like there is an issue with a plugin we are using.
The log file is attached below.

Could someone help us with this error please ?

[crash.log](https://github.com/hashicorp/packer-plugin-azure/files/14997913/crash.log)

JenGoldstrich:
Previously we weren't instantiating the managed image object inside the template factory whenever a build used a SIG image as a source, (i.e. had a shared_image_gallery block set), this would cause a panic when setting a disk encryption set ID, and as of v2.0.5. it would also cause a similar panic when trying to set the encryption type.  To verify this I went back to version 2.0.0 and also reproduced a panic when setting a disk encryption set ID with a `shared_image_gallery` block as the source image

To fix this I just added nullchecks and instantiate where relevant, I ran acceptance tests on this PR

Closes #406

### DIFF
--- a/builder/azure/arm/template_factory_test.go
+++ b/builder/azure/arm/template_factory_test.go
@@ -748,6 +748,41 @@ func TestTrustedLaunch01(t *testing.T) {
 	approvaltests.VerifyJSONStruct(t, deployment.Properties.Template)
 }
 
+func TestSigSourcedWithDiskEncryptionSet(t *testing.T) {
+	m := map[string]interface{}{
+		"build_resource_group_name": "rg_name",
+		"disk_encryption_set_id":    "my_id",
+		"shared_image_gallery_destination": map[string]interface{}{
+			"subscription":   "subscription",
+			"resource_group": "rg_name",
+			"gallery_name":   "gallery_name",
+			"image_name":     "image_name",
+			"image_version":  "2024.01.02",
+		},
+		"shared_image_gallery": map[string]interface{}{
+			"subscription":   "subscription",
+			"resource_group": "rg_name",
+			"gallery_name":   "gallery_name",
+			"image_name":     "image_name",
+			"image_version":  "2024.01.01",
+		},
+		"os_type":      constants.Target_Linux,
+		"communicator": "none",
+	}
+
+	var c Config
+	_, err := c.Prepare(m, getPackerConfiguration(), getPackerSSHPasswordCommunicatorConfiguration())
+	if err != nil {
+		t.Fatal(err)
+	}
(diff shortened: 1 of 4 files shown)


builder/azure/arm/template_factory_test.go
```diff
@@ -748,6 +748,41 @@ func TestTrustedLaunch01(t *testing.T) {
 	approvaltests.VerifyJSONStruct(t, deployment.Properties.Template)
 }
 
+func TestSigSourcedWithDiskEncryptionSet(t *testing.T) {
```
JenGoldstrich, line 751: These two tests panic without the changes to template_builder.go to validate the fix

lbajolet-hashicorp (approved):
LGTM!

PR #410, opened 2024-04-17
fix: [arm builder] storage blob VHD deletion, cleanup post build deletion

### LINKED ISSUE #273: NIC not disassociated before deletion
#### Overview of the Issue

When running Packer with the Azure plugin, it doesn't disassociate the NIC before deleting the VM, so it crashes (see log below).

#### Reproduction Steps

Run Packer to create an Azure VM image. The buildfile is here: https://github.com/actions/runner-images/blob/main/images/linux/ubuntu2004.json

### Plugin and Packer version

Packer: 1.8.5
Packer Azure plugin: 1.3.1

https://github.com/actions/runner-images/blob/main/images/linux/ubuntu2004.json

The Packer process runs on `ubuntu-latest` in an Azure DevOps Pipeline agent. The VM image Packer is creating is based on Ubuntu 20.04 LTS (https://github.com/actions/runner-images)

```
==> azure-arm: Adding to deletion queue -> Microsoft.Network/networkInterfaces : 'pkrniho2hi13cqd'
==> azure-arm: Attempting deletion -> Microsoft.Network/networkInterfaces : 'pkrniho2hi13cqd'
==> azure-arm: Waiting for deletion of all resources...
==> azure-arm: Attempting deletion -> Microsoft.Compute/virtualMachines : 'pkrvmho2hi13cqd'
==> azure-arm: Error deleting resource. Will retry.
==> azure-arm: Name: pkrniho2hi13cqd
==> azure-arm: Error: network.InterfacesClient#Delete: Failure sending request: StatusCode=400 -- Original Error: Code="NicInUse" Message="Network Interface /subscriptions/[ID]/resourceGroups/[RG]/providers/Microsoft.Network/networkInterfaces/pkrniho2hi13cqd is used by existing resource /subscriptions/[ID]/resourceGroups/[RG]/providers/Microsoft.Compute/virtualMachines/pkrvmho2hi13cqd. In order to delete the network interface, it must be dissociated from the resource. To learn more, see aka.ms/deletenic." Details=[]
==> azure-arm:
==> azure-arm: Attempting deletion -> Microsoft.Network/networkInterfaces : 'pkrniho2hi13cqd'
==> azure-arm: Error deleting resource. Will retry.
==> azure-arm: Name: pkrniho2hi13cqd
==> azure-arm: Error: network.InterfacesClient#Delete: Failure sending request: StatusCode=400 -- Original Error: Code="NicInUse" Message="Network Interface /subscriptions/[ID]/resourceGroups/[RG]/providers/Microsoft.Network/networkInterfaces/pkrniho2hi13cqd is used by existing resource /subscriptions/[ID]/resourceGroups/[RG]/providers/Microsoft.Compute/virtualMachines/pkrvmho2hi13cqd. In order to delete the network interface, it must be dissociated from the resource. To learn more, see aka.ms/deletenic." Details=[]
==> azure-arm:
==> azure-arm: Attempting deletion -> Microsoft.Network/networkInterfaces : 'pkrniho2hi13cqd'
==> azure-arm: Error deleting resource. Will retry.
==> azure-arm: Name: pkrniho2hi13cqd
==> azure-arm: Error: network.InterfacesClient#Delete: Failure sending request: StatusCode=400 -- Original Error: Code="NicInUse" Message="Network Interface /subscriptions/[ID]/resourceGroups/[RG]/providers/Microsoft.Network/networkInterfaces/pkrniho2hi13cqd is used by existing resource /subscriptions/[ID]/resourceGroups/[RG]/providers/Microsoft.Compute/virtualMachines/pkrvmho2hi13cqd. In order to delete the network interface, it must be dissociated from the resource. To learn more, see aka.ms/deletenic." Details=[]
==> azure-arm:
==> azure-arm: Attempting deletion -> Microsoft.Network/networkInterfaces : 'pkrniho2hi13cqd'
==> azure-arm:  Deleting -> Microsoft.Compute/disks : '/subscriptions/[ID]/resourceGroups/[RG]/providers/Microsoft.Compute/disks/pkrosho2hi13cqd'
==> azure-arm: Removing the created Deployment object: 'pkrdpho2hi13cqd'
==> azure-arm: 
==> azure-arm: The resource group was not created by Packer, not deleting ...
```

### LINKED ISSUE #409: v2.1.0: Storage account deletion URI invalid, wrong parameters sent
In v2.1.0 we upgrdaded the version of Giovanni we use https://github.com/tombuildsstuff/giovanni which is used to delete storage blobs for OS/additional disks for VHD builds, the parameters for the delete method https://github.com/tombuildsstuff/giovanni/blob/41801b9419065c1e6a6f369a7f03c9f51bfffe6f/storage/2020-08-04/blob/blobs/delete.go#L28 changed to no longer pass in storage account, however I misread this during development, and thought the container was removed, instead storage account now comes from the URL given to the base client.

This leads to some orphaned disks on VHD builds.

JenGoldstrich:
This PR has three main purposes

- Fix #409 by passing in storage account name to client as expected
- Fix #273 by deleting the VM/Network Interface before deleting other networking resources, these errors were harmless and didn't actually lead to orphaned resources, but its noisy and unnecessary
- Make it possible to unit test these new changes with substantial refactoring to step_deploy_template's CleanUp step, moving several of the resource deletions into their own functions.

Closes #409
Closes #273

Before this PR when building a VHD image with additional disks that should be deleted, you can see extra print statements at the bottom of 2, and my storage account as well as the failed additional disk deletion, and NIC repeated retry errors.

```
==> azure-arm:
==> azure-arm: Deleting Virtual Machine deployment and its attached resources...
==> azure-arm: Adding to deletion queue -> Microsoft.Compute/virtualMachines : 'replaced'
==> azure-arm: Adding to deletion queue -> Microsoft.Network/networkInterfaces : 'replaced'
==> azure-arm: Adding to deletion queue -> Microsoft.Network/virtualNetworks : 'replaced'
==> azure-arm: Adding to deletion queue -> Microsoft.Network/publicIPAddresses : 'replaced'
==> azure-arm: Waiting for deletion of all resources...
==> azure-arm: Attempting deletion -> Microsoft.Compute/virtualMachines : 'replaced'==> azure-arm: Attempting deletion -> Microsoft.Network/publicIPAddresses : 'replaced'
==> azure-arm: Attempting deletion -> Microsoft.Network/virtualNetworks : 'replaced'==> azure-arm: Attempting deletion -> Microsoft.Network/networkInterfaces : 'replaced'
==> azure-arm: Couldn't delete Microsoft.Network/publicIPAddresses resource. Will retry.
==> azure-arm: Name: replaced
==> azure-arm: Couldn't delete Microsoft.Network/virtualNetworks resource. Will retry.
==> azure-arm: Name: replaced
==> azure-arm: Attempting deletion -> Microsoft.Network/publicIPAddresses : 'replaced'
==> azure-arm: Attempting deletion -> Microsoft.Network/virtualNetworks : 'replaced'==> azure-arm: Skipping deletion -> image : 'https://removedstorageaccount.blob.core.windows.net/images/removed.vhd' since 'keep_os_disk' is set to true
==> azure-arm:  Deleting Additional Disk -> 1: 'https://removedstorageaccount.blob.core.windows.net/images/pkrdisk-1.vhd'
==> azure-arm: Failed to delete the managed Additional Disk!
==> azure-arm:  Deleting Additional Disk -> 2: 'https://removedstorageaccount.blob.core.windows.net/images/pkrdisk-2.vhd'
==> azure-arm: Failed to delete the managed Additional Disk!
==> azure-arm: Removing the created Deployment object: 'pkrdplafewlqjyv'
==> azure-arm:
==> azure-arm: The resource group was not created by Packer, not deleting ...
==> azure-arm: https://removedstorageaccount.blob.core.windows.net/
==> azure-arm: 2
Build 'azure-arm' finished after 3 minutes 6 seconds.
```

After

```
==> azure-arm: Deleting Virtual Machine deployment and its attached resources...
==> azure-arm: Deleted -> Microsoft.Compute/virtualMachines : 'pkrvmq58v6zr4b1'
==> azure-arm: Deleted -> Microsoft.Network/networkInterfaces : 'pkrniq58v6zr4b1'
==> azure-arm: Deleted -> Microsoft.Network/publicIPAddresses : 'pkripq58v6zr4b1'
==> azure-arm: Deleted -> Microsoft.Network/virtualNetworks : 'pkrvnq58v6zr4b1'
==> azure-arm: Skipping deletion -> image : 'https://removed.blob.core.windows.net/images/removed.vhd' since 'keep_os_disk' is set to true
==> azure-arm:  Deleted Additional Disk -> 1: 'https://removed.blob.core.windows.net/images/removed-1.vhd'
==> azure-arm:  Deleted Additional Disk -> 2: 'https://removed.blob.core.windows.net/images/removed-2.vhd'
==> azure-arm: Removing the created Deployment object: 'pkrdpq58v6zr4b1'
==> azure-arm:
==> azure-arm: The resource group was not created by Packer, not deleting ...
Build 'azure-arm' finished after 3 minutes 1 second.
```

### DIFF
--- a/builder/azure/arm/azure_client.go
+++ b/builder/azure/arm/azure_client.go
@@ -77,7 +77,7 @@ func errorCapture(client *AzureClient) client.ResponseMiddleware {
 }
 
 // Returns an Azure Client used for the Azure Resource Manager
-func NewAzureClient(ctx context.Context, isVHDBuild bool, cloud *environments.Environment, sharedGalleryTimeout time.Duration, pollingDuration time.Duration, authOptions commonclient.AzureAuthOptions) (*AzureClient, error) {
+func NewAzureClient(ctx context.Context, storageAccountName string, cloud *environments.Environment, sharedGalleryTimeout time.Duration, pollingDuration time.Duration, authOptions commonclient.AzureAuthOptions) (*AzureClient, error) {
 
 	var azureClient = &AzureClient{}
 
@@ -230,13 +230,12 @@ func NewAzureClient(ctx context.Context, isVHDBuild bool, cloud *environments.En
 	azureClient.GalleryImagesClient = *galleryImagesClient
 
 	// We only need the Blob Client to delete the OS VHD during VHD builds
-	if isVHDBuild {
+	if storageAccountName != "" {
 		storageAccountAuthorizer, err := commonclient.BuildStorageAuthorizer(ctx, authOptions, *cloud)
 		if err != nil {
 			return nil, err
 		}
-
-		blobClient, err := giovanniBlobStorageSDK.NewWithBaseUri(cloud.Storage.Name())
+		blobClient, err := giovanniBlobStorageSDK.NewWithBaseUri(fmt.Sprintf("https://%s.blob.core.windows.net", storageAccountName))
 		if err != nil {
 			return nil, err
 		}
--- a/builder/azure/arm/step_deploy_template.go
+++ b/builder/azure/arm/step_deploy_template.go
@@ -7,6 +7,7 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"log"
 	"net/url"
 	"strings"
 	"sync"
@@ -30,21 +31,28 @@ type DeploymentTemplateType int
 const (
 	VirtualMachineTemplate DeploymentTemplateType = iota
 	KeyVaultTemplate
+	VMResourceType               = "Microsoft.Compute/virtualMachines"
+	NetworkInterfaceResourceType = "Microsoft.Network/networkInterfaces"
+	KeyVaultResourceType         = "Microsoft.KeyVault/vaults"
 )
 
 type StepDeployTemplate struct {
-	client           *AzureClient
-	deploy           func(ctx context.Context, subscriptionId string, resourceGroupName string, deploymentName string) error
-	delete           func(ctx context.Context, subscriptionId, deploymentName, resourceGroupName string) error
-	disk             func(ctx context.Context, subscriptionId string, resourceGroupName string, computeName string) (string, string, error)
-	deleteDisk       func(ctx context.Context, imageName string, resourceGroupName string, isManagedDisk bool, subscriptionId string, storageAccountName string) error
-	deleteDeployment func(ctx context.Context, state multistep.StateBag) error
-	say              func(message string)
-	error            func(e error)
-	config           *Config
-	factory          templateFactoryFunc
-	name             string
-	templateType     DeploymentTemplateType
+	client                   *AzureClient
+	deploy                   func(ctx context.Context, subscriptionId string, resourceGroupName string, deploymentName string) error
+	deleteDetatchedResources func(ctx context.Context, subscriptionId string, resourceGroupName string, resources map[string]string)
+	getDisk                  func(ctx context.Context, subscriptionId string, resourceGroupName string, computeName string) (string, string, error)
+	deleteDisk               func(ctx context.Context, imageName string, resourceGroupName string, isManagedDisk bool, subscriptionId string, storageAccountName string) error
+	deleteVM                 func(ctx context.Context, virtualMachineId virtualmachines.VirtualMachineId) error
+	deleteNic                func(ctx context.Context, networkInterfacesId commonids.NetworkInterfaceId) error
+	deleteDeployment         func(ctx context.Context, state multistep.StateBag) error
+	deleteKV                 func(ctx context.Context, id commonids.KeyVaultId) error
+	listDeploymentOps        func(ctx context.Context, id deploymentoperations.ResourceGroupDeploymentId) ([]deploymentoperations.DeploymentOperation, error)
+	say                      func(message string)
+	error                    func(e error)
+	config                   *Config
+	factory                  templateFactoryFunc
+	name                     string
+	templateType             DeploymentTemplateType
 }
 
 func NewStepDeployTemplate(client *AzureClient, ui packersdk.Ui, config *Config, deploymentName string, factory templateFactoryFunc, templateType DeploymentTemplateType) *StepDeployTemplate {
@@ -59,10 +67,14 @@ func NewStepDeployTemplate(client *AzureClient, ui packersdk.Ui, config *Config,
 	}
 
 	step.deploy = step.deployTemplate
-	step.delete = step.deleteDeploymentResources
-	step.disk = step.getImageDetails
+	step.getDisk = step.getImageDetails
 	step.deleteDisk = step.deleteImage
+	step.deleteDetatchedResources = step.deleteDetatchedResourcesWithQueue
 	step.deleteDeployment = step.deleteDeploymentObject
+	step.deleteNic = step.deleteNetworkInterface
+	step.deleteVM = step.deleteVirtualMachine
+	step.deleteKV = step.deleteKeyVault
+	step.listDeploymentOps = step.listDeploymentOperations
 	return step
 }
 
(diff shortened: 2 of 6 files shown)


builder/azure/arm/azure_client.go
```diff
@@ -230,13 +230,12 @@ func NewAzureClient(ctx context.Context, isVHDBuild bool, cloud *environments.En
...
 
 	// We only need the Blob Client to delete the OS VHD during VHD builds
-	if isVHDBuild {
+	if storageAccountName != "" {
 		storageAccountAuthorizer, err := commonclient.BuildStorageAuthorizer(ctx, authOptions, *cloud)
 		if err != nil {
 			return nil, err
 		}
-
-		blobClient, err := giovanniBlobStorageSDK.NewWithBaseUri(cloud.Storage.Name())
```
JenGoldstrich, line 239: This actually was just passing in "AzureStorage" as a string, however the client never invoked this so we never ran into the issue since Delete blobs endpoint previously accepted storage account, now we had to give it this URL.

See https://learn.microsoft.com/en-us/azure/storage/common/storage-account-overview#standard-endpoints for more info

builder/azure/arm/step_deploy_template.go
```diff
@@ -298,21 +286,48 @@ func (s *StepDeployTemplate) deleteDeploymentResources(ctx context.Context, subs
...
 			continue
 		}
-
 		resourceName := *deploymentOperation.Properties.TargetResource.ResourceName
 		resourceType := *deploymentOperation.Properties.TargetResource.ResourceType
 
-		s.say(fmt.Sprintf("Adding to deletion queue -> %s : '%s'", resourceType, resourceName))
-		resources[resourceType] = resourceName
-
+		// Create vm and NIC ID to use them to delete later
```
JenGoldstrich, line 299: One small quirk of this change is that previously we would call Delete VM/NIC even if it failed once, we don't expect this to happen but technically we would keep retrying that call. So far I've only seen the Public IP and VNET deletions would fail due to a still existing network interface. So it shouldn't be a problem but just figured I'd call it out and see if the reviewer of this PR has any thoughts :D

JenGoldstrich: One option is just to ship it as is, and then if we here reports from users that they're seeing orphaned VMs we can quickly re-add it to the deletion queue in the event that it fails, basically falling back on the old mechanism

builder/azure/arm/step_deploy_template.go
```diff
@@ -143,7 +143,7 @@ func (s *StepDeployTemplate) Cleanup(state multistep.StateBag) {
 
 			err := s.deleteImage(ctx, additionaldisk, resourceGroupName, (isManagedDisk || isSIGImage), subscriptionId, armStorageAccountName)
 			if err != nil {
-				s.say("Failed to delete the managed Additional Disk!")
+				s.say(fmt.Sprintf("Failed to delete the managed Additional Disk! %s", err))
```
JenGoldstrich, line 146: We weren't actually logging the error here 🤦‍♀️

go.mod
```diff
@@ -22,7 +22,7 @@ require (
 	github.com/hashicorp/go-azure-sdk/resource-manager v0.20240411.1145857
 	github.com/hashicorp/go-azure-sdk/sdk v0.20240411.1145857
 	github.com/mitchellh/go-homedir v1.1.0
-	github.com/tombuildsstuff/giovanni v0.25.3
+	github.com/tombuildsstuff/giovanni v0.26.1
```
JenGoldstrich, line 25: This update isn't required, I was just seeing when the change was introduced and figured I'd upgrade while testing

builder/azure/arm/step_deploy_template.go
```diff
@@ -298,21 +286,48 @@ func (s *StepDeployTemplate) deleteDeploymentResources(ctx context.Context, subs
...
+			s.say(fmt.Sprintf("Adding to deletion queue -> %s : '%s'", resourceType, resourceName))
+			resources[resourceType] = resourceName
+		}
 	}
 
+	// Always delete VM first, then delete NIC, this way we avoid conflicts on deletion
+	if vmID != nil {
+		s.say(fmt.Sprintf("Attempting deletion -> %s : '%s'", "Microsoft.Compute/virtualMachines", vmID.VirtualMachineName))
+		if err := s.client.VirtualMachinesClient.DeleteThenPoll(pollingContext, *vmID, virtualmachines.DefaultDeleteOperationOptions()); err != nil {
+			s.reportIfError(err, vmID.VirtualMachineName)
```
lbajolet-hashicorp, line 317: I wonder looking at this, why do we have a function called `reportIfError` if we conditionally call it only in case of error?

builder/azure/arm/step_deploy_template.go
```diff
@@ -335,6 +350,8 @@ func (s *StepDeployTemplate) deleteDeploymentResources(ctx context.Context, subs
 					s.say(fmt.Sprintf("Couldn't delete %s resource. Will retry.\n"+
```
lbajolet-hashicorp, line 350: This should probably become a `log.Printf` so we don't flood users with non-actionable error messages
```suggestion
					log.Printf("[INFO] Couldn't delete resource %s.%s, will retry", resourceType, resourceName)
```

builder/azure/arm/step_deploy_template.go
```diff
@@ -298,21 +286,48 @@ func (s *StepDeployTemplate) deleteDeploymentResources(ctx context.Context, subs
...
+		default:
+			s.say(fmt.Sprintf("Adding to deletion queue -> %s : '%s'", resourceType, resourceName))
+			resources[resourceType] = resourceName
+		}
 	}
 
+	// Always delete VM first, then delete NIC, this way we avoid conflicts on deletion
+	if vmID != nil {
+		s.say(fmt.Sprintf("Attempting deletion -> %s : '%s'", "Microsoft.Compute/virtualMachines", vmID.VirtualMachineName))
+		if err := s.client.VirtualMachinesClient.DeleteThenPoll(pollingContext, *vmID, virtualmachines.DefaultDeleteOperationOptions()); err != nil {
```
lbajolet-hashicorp, line 316: To be clear, does `DeleteThenPoll` gets retried if moved out of the loop below? I can see a world where this fails for many reasons that may be transient (temporary connection loss, busy servers, endpoint failure for whatever reason, etc.) , failing completely feels suboptimal if so.

builder/azure/arm/step_deploy_template.go
```diff
@@ -335,6 +350,8 @@ func (s *StepDeployTemplate) deleteDeploymentResources(ctx context.Context, subs
 					s.say(fmt.Sprintf("Couldn't delete %s resource. Will retry.\n"+
 						"Name: %s",
 						resourceType, resourceName))
+				} else {
+					s.say(fmt.Sprintf("Finished deleting -> %s : '%s'", resourceType, resourceName))
```
lbajolet-hashicorp, line 354: In terms of UX, I wonder if it wouldn't be clearer to keep the successes/failures in a list and display them all in one message once the processing is done? This would reduce the amount of noise produced by the plugin during this step I would think

builder/azure/arm/step_deploy_template.go
JenGoldstrich: It does not, it may be worth adding them to their own retry blocks, but separate from each other, ideally we don't leave behind VMs which I believe is why everything was in a retry block, however this lead to a lot of confusion and frustration from users who think the errors are a bit annoying.

Do you think it'd be worthwhile to create two extra retry blocks, one for VMs and one for Network interfaces, to ensure they both get deleted, even if the API flakes (tbh I've never seen this happen)

builder/azure/arm/step_deploy_template.go
JenGoldstrich: @lbajolet-hashicorp in several places we don't call if it there is an error, I was just calling the same function to re-use the same warning message

builder/azure/arm/step_deploy_template.go
JenGoldstrich: That's fair, we don't expect this error to occur anymore unless something is actually going weird though, but since we're gonna retry it seems worthwhile to add that into logs instead.

builder/azure/arm/step_deploy_template.go
```diff
@@ -191,9 +274,6 @@ func (s *StepDeployTemplate) getImageDetails(ctx context.Context, subscriptionId
 	defer cancel()
 	vmID := virtualmachines.NewVirtualMachineID(subscriptionId, resourceGroupName, computeName)
 	vm, err := s.client.VirtualMachinesClient.Get(pollingContext, vmID, virtualmachines.DefaultGetOperationOptions())
-	if err != nil {
```
JenGoldstrich, line 194: This was invalid code, the lower error was unreachable

builder/azure/arm/step_deploy_template.go
JenGoldstrich: I would prefer to not do that for now, with my refactored logic we'll just error saying that the VM, NIC, or resource failed, if the VM or NIC failed, the user will go manually delete it and see its associated resource, if we get an issue from users where this is confusing we can change this in the future

builder/azure/arm/step_deploy_template.go
JenGoldstrich: We now retry this

builder/azure/arm/step_deploy_template.go
```diff
@@ -279,82 +324,171 @@ func (s *StepDeployTemplate) deleteImage(ctx context.Context, imageName string,
 	if len(xs) < 3 {
 		return errors.New("Unable to parse path of image " + imageName)
 	}
-	_, err = s.client.GiovanniBlobClient.Delete(pollingContext, storageAccountName, blobName, giovanniBlobStorageSDK.DeleteInput{})
+	_, err = s.client.GiovanniBlobClient.Delete(pollingContext, "images", blobName, giovanniBlobStorageSDK.DeleteInput{})
 	return err
 }
 
-func (s *StepDeployTemplate) deleteDeploymentResources(ctx context.Context, subscriptionId, deploymentName, resourceGroupName string) error {
+func (s *StepDeployTemplate) deleteVirtualMachine(ctx context.Context, vmID virtualmachines.VirtualMachineId) error {
```
JenGoldstrich, line 331: While this requires a lot more code, I think its better to explicitly break apart these steps, the current function is really hard to debug when stuff goes wrong

builder/azure/arm/step_deploy_template.go
```diff
@@ -93,11 +105,38 @@ func (s *StepDeployTemplate) Cleanup(state multistep.StateBag) {
...
+				}
+				err := s.deleteKV(ctx, kvID)
+				if err != nil {
+					s.reportResourceDeletionFailure(err, resourceGroupName)
+				}
+				return
+			}
 		}
 
 	} else {
```
lbajolet-hashicorp, line 142: This `else` feels like it could be removed if we return at the end of that `if`, since the next action is essentially returning anyway

builder/azure/arm/step_deploy_template.go
```diff
@@ -108,13 +147,55 @@ func (s *StepDeployTemplate) Cleanup(state multistep.StateBag) {
...
+				return
+			}
+		}
+		if networkInterfaceID != nil {
+			err := s.deleteNic(ctx, *networkInterfaceID)
+			if err != nil {
+				return
+			}
+		}
+		err = s.deleteNetworkResources(ctx, subscriptionId, resourceGroupName, resources)
```
lbajolet-hashicorp, line 196: I see you added a `deleteNetworkResourcesWithQueue` function, was that the one you meant to call here? What's the difference between those two?

lbajolet-hashicorp: Also come to think of it, I see we remove the NIC before executing this function, but I would think a NIC is part of the network resources, not saying we should move the NIC deletion to this function necessarily since I believe having it deleted beforehand is the purpose of this PR, but maybe the function should be renamed

builder/azure/arm/step_deploy_template.go
lbajolet-hashicorp: Most of the code looks similar, I'd think we can factorise it a bit more in order to reduce the amount of duplicated code, but that can be done later, no need to make this PR more complex today

builder/azure/arm/step_deploy_template.go
```diff
@@ -279,82 +324,141 @@ func (s *StepDeployTemplate) deleteImage(ctx context.Context, imageName string,
...
+		if err != nil {
+			log.Printf("[INFO] Couldn't delete resource %s.%s, will retry", resourceType, resourceName)
+		}
+		return err
+	})
+	if err != nil {
+		s.reportResourceDeletionFailure(err, resourceName)
+	}
+	return err
+
```
lbajolet-hashicorp, line 352: Tiniest of nits: extra blank line here

builder/azure/arm/step_deploy_template.go
```diff
@@ -279,82 +324,141 @@ func (s *StepDeployTemplate) deleteImage(ctx context.Context, imageName string,
...
 		go func(resourceType, resourceName string) {
 			defer wg.Done()
-			retryConfig := retry.Config{
-				Tries:      10,
-				RetryDelay: (&retry.Backoff{InitialBackoff: 5 * time.Second, MaxBackoff: 60 * time.Second, Multiplier: 1.5}).Linear,
-			}
 
-			err = retryConfig.Run(ctx, func(ctx context.Context) error {
-				s.say(fmt.Sprintf("Attempting deletion -> %s : '%s'", resourceType, resourceName))
+			err := retryConfig.Run(ctx, func(ctx context.Context) error {
```
lbajolet-hashicorp, line 415: Any reason why we don't use `retryDeletion` for those resources?

lbajolet-hashicorp (approved):
Left a few last nits, but overall this LGTM!

builder/azure/arm/step_deploy_template.go
JenGoldstrich: Nope, good call out, I moved this to using that!

PR #416, opened 2024-05-16
Add support for setting public IP sku to ARM builder

### LINKED ISSUE #292: azure-arm: Add `public ip sku` parameter
Please search the existing issues for relevant feature requests, and use the
reaction feature
(https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/)
to add upvotes to pre-existing requests.

#### Community Note

#### Description

Under certain configurations, the packer creates a Public IP to be used by the virtual machine but is always a basic SKU, and does not exist a parameter to define a different SKU.
The request is to have that possibility.

#### Use Case(s)

As part of the process, some tools are installed and some information is retrieved from an IP-restricted site. To be able to configure an appropriate allow-list is necessary to control the outbound IP, and for that purpose, I have a NAT Gateway associated with the subnet used by Packer.
The whole process fails because Azure does allow the creation of any other public IP with the Basic Sku in the same subnet as NAT Gateway.

#### Potential configuration

HCL
```
source "azure-arm" "example" {
  ...
  public_ip_sku = "Standard"
  ...
}
```

JenGoldstrich:
This PR adds support for a `public_ip_sku` parameter, in March 2025 Microsoft is removing the ability to create Basic SKU public IP addresses, there are certain security aspects of this that do not really apply to the short lived IP that is created to connect to the build VM.  Due to the many number of potential network configurations we want to release support for this early and encourage users to try out this functionality, so any unexpected bugs are caught long before the removal of basic IPs.  This will minimize impact to our users.

This PR also throws a new warning to inform users of the upcoming deprecation, and suggest they try this field

Closes #292


builder/azure/arm/config.go
```diff
@@ -623,6 +624,12 @@ type Config struct {
 	// Requires enabling encryption at host in the Subscription read more [here](https://learn.microsoft.com/en-us/azure/virtual-machines/disks-enable-host-based-encryption-portal?tabs=azure-powershell)
 	EncryptionAtHost *bool `mapstructure:"encryption_at_host" required:"false"`
 
+	// Specifies the SKU to use in the Public IP Address created to connect to the build Virtual machine.
+	// Can only be `Basic`, or `Standard`. Currently basic is the default if none is set.
+	// On 31 March 2025 Azure will remove the ability to create `Basic` SKU public IPs, before then the plugin will be updated to create a Standard public IP by default
+	// You can read more about public IP skus [here](https://learn.microsoft.com/en-us/azure/virtual-network/ip-services/public-ip-addresses#sku)
```
nywilken, line 630: Slight reword to match the wording for similar SKU options and replace "read" with something a bit more inclusive. 
```suggestion
	// Specify the Public IP Address SKU for the public IP used to connect to the build Virtual machine.
	// Valid values are `Basic` and `Standard`. The default value is `Basic`.
	// On 31 March 2025 Azure will remove the ability to create `Basic` SKU public IPs, before then the plugin will be updated to create a Standard public IP by default
	// You can learn more about public IP SKUs [here](https://learn.microsoft.com/en-us/azure/virtual-network/ip-services/public-ip-addresses#sku)
```

On a side note, I see the SKUs are capitalized should we be checking for this? Or is the upstream API agnostic to casing?

builder/azure/arm/builder.go
```diff
@@ -131,6 +132,10 @@ func (b *Builder) Run(ctx context.Context, ui packersdk.Ui, hook packersdk.Hook)
 	if b.config.ClientConfig.ObjectID == "" && b.config.OSType != constants.Target_Linux {
 		return nil, fmt.Errorf("could not determine the ObjectID for the user, which is required for Windows builds")
 	}
+	publicIPWarning := "On 31 March 2025 Azure is disabling the ability to create public IP Addresses with the Basic SKU, this builder has always created a basic SKU public IP Address to connect to the build VM.  We reccomend setting the `public_ip_sku` field to `Standard`, this will make sure that your build is created with a standard public IPs, and help ensure any issues with this change are caught before the removal of Basic SKU public IPs. At that point this plugin will be updated to default to Standard IP skus. You can read more about this in the official Azure announcement https://azure.microsoft.com/en-us/updates/upgrade-to-standard-sku-public-ip-addresses-in-azure-by-30-september-2025-basic-sku-will-be-retired/"
```
nywilken, line 135: Slight reword based off my understanding of the warning. 
```suggestion
	publicIPWarning := "On March 31, 2025, Azure will no longer allow the creation of public IP addresses with a Basic SKU, which is the default SKU type for this builder. We recommend testing your build using the standard public IP SKU by setting the `public_ip_sku` field to `Standard`. This builder will update its default value from `Basic` to `Standard` in a future release closer to Azure’s removal date. You can learn more about this change in the official Azure announcement https://azure.microsoft.com/en-us/updates/upgrade-to-standard-sku-public-ip-addresses-in-azure-by-30-september-2025-basic-sku-will-be-retired/."
```

builder/azure/arm/builder.go
```diff
@@ -131,6 +132,10 @@ func (b *Builder) Run(ctx context.Context, ui packersdk.Ui, hook packersdk.Hook)
...
+	if b.config.PublicIpSKU == "" || b.config.PublicIpSKU == string(publicipaddresses.PublicIPAddressSkuNameBasic) {
```
nywilken, line 136: This warning will output for every user once this is released. If the intent it to alert users starting now, we need to update the messaging above to inform the user on how to remove this warning. Which appears to be just using a `Standard` SKU. 

If a user is unable to use a Standard SKU we will keep alerting if they set the field to the only value they can use `Basic`.  I wonder if it makes sense to warn if the value is explicitly set to Basic and in the next two releases update the warning to include those users who are not setting the value.

nywilken (commented):
I left a couple of questions and small suggestions. But this otherwise looks good.

builder/azure/arm/builder.go
JenGoldstrich: I would rather warn always, and if users can not upgrade now, they will open issues which will reveal the reasons why they can not upgrade, and allow us to fix it, we want users to upgrade as soon as possible so 

1.) We can find out any missing functionality before the deadline
2.) Users are on the new IP sku long before the deadline

JenGoldstrich:
@nywilken I have merged your changes, run make generate, and made public IP sku case insensitive for the user, lemme know what you think!

PR #429, opened 2024-06-20
Don't hardcode go version in acceptance tests

JenGoldstrich:
We hardcore the go version in the acceptance tests currently to go 1.19, which causes a failure currently.

But in general we shouldn't hardcore this version


lbajolet-hashicorp (approved):
LGTM!

Issue #430, opened 2024-07-10
Update documentation for target_regions block


dhananjaipai:
#### Overview of the Issue

The [documentation](https://developer.hashicorp.com/packer/integrations/hashicorp/azure/latest/components/builder/arm) for `target_regions` does not specify the expected inputs in the block.
It just shows
```
target_region ([]TargetRegion) - A target region to store the image version in. ...
```
with no reference to what `[]TargetRegion` is.

#### Reproduction Steps

Try setting `target_region` based on official [docs](https://learn.microsoft.com/en-us/rest/api/compute/gallery-image-versions/create-or-update?view=rest-compute-2024-03-01&tabs=HTTP#targetregion) and it does not work

The PR #387 seems to use `replicas` instead of `regionalReplicaCount` or `regional_replica_count` as expected, and there is no reference to this in the documentation.

### Plugin and Packer version

From `packer version` 1.11.1

JenGoldstrich:
Hey @dhananjaipai thanks for raising this issue

I have addressed the documentation issue and you can now see the target regions docs [here](https://developer.hashicorp.com/packer/integrations/hashicorp/azure/latest/components/builder/arm#target-regions)

eugeneswalker:
> Hey @dhananjaipai thanks for raising this issue
> 
> I have addressed the documentation issue and you can now see the target regions docs [here](https://developer.hashicorp.com/packer/integrations/hashicorp/azure/latest/components/builder/arm#target-regions)

If you search that page for "target_region" (singular, not plural) you should see there is still example HCL2 showing 
```
shared_image_gallery_destination {
    ...
    target_region {
      name = "regionA"
    }
    target_region {
      name = "regionB"
    }
    target_region {
      name = "regionC"
    }
}
```

Does this need updating too?

JenGoldstrich:
@eugeneswalker I am not understanding what you are suggesting should be updated, can you please open a new issue describing what you think needs to be updated here and we'll address it as we're able

eugeneswalker:
>
> #### Reproduction Steps
> 
> Try setting `target_region` based on official [docs](https://learn.microsoft.com/en-us/rest/api/compute/gallery-image-versions/create-or-update?view=rest-compute-2024-03-01&tabs=HTTP#targetregion) and it does not work

This is still the case with the documentation linked in the first sentence of this issue:
* https://developer.hashicorp.com/packer/integrations/hashicorp/azure/latest/components/builder/arm
[...]

PR #437, opened 2024-08-22
Add docs for `target_region` replicas field, make error message for it clearer

JenGoldstrich:
There's been some confusion around the target_region block and its replicas field, this PR fixes an issue where there we weren't including the specific information for the fields the target_region block in the docs.  I also modified the error message to make this easier to understand

closes https://github.com/hashicorp/packer-plugin-azure/issues/430


nywilken:
After merging I suggest running the manual notify integration release job to trigger a re-injest of the plugins docs for the 2.1.7 release.

PR #438, opened 2024-08-22
Fix `replicated_regions` not using `shared_image_gallery_replica_count` for build location when its specified in the list

JenGoldstrich:
The logic that handles the backwards compatibility for creating the new Target Regions block from the replicated regions list doesn't add the replica count to the default region.  This causes the default region to always have a replica count of 1, this is currently a bit iffy to unit test and should probably be unwound into its own step still, but I don't think we should leave this functionality broken in the mean time

### DIFF
--- a/builder/azure/arm/builder.go
+++ b/builder/azure/arm/builder.go
@@ -261,9 +261,16 @@ func (b *Builder) Run(ctx context.Context, ui packersdk.Ui, hook packersdk.Hook)
 			for _, region := range b.config.SharedGalleryDestination.SigDestinationReplicationRegions {
 				region := normalizeAzureRegion(region)
 				if strings.EqualFold(region, buildLocation) {
-					// backwards compatibility DiskEncryptionSetId was set on the global config not on the target region.
-					// Users using target_region blocks are responsible for setting the DES within the block
-					normalizedRegions = append(normalizedRegions, TargetRegion{Name: region, DiskEncryptionSetId: b.config.DiskEncryptionSetId})
+					normalizedRegions = append(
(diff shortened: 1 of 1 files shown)


JenGoldstrich:
I explained this on slack but just for other's context the issue here is that we don't ever leave it unset

Firstly https://github.com/hashicorp/packer-plugin-azure/blob/4d2a02095e1785c6df199f466ea8c70d42449f2e/builder/azure/arm/builder.go#L272-L278 we set this for non-build location target regions currently and in https://github.com/hashicorp/packer-plugin-azure/blob/4d2a02095e1785c6df199f466ea8c70d42449f2e/builder/azure/arm/step_publish_to_shared_image_gallery.go#L113-L117 publish step if its not set ( 0 in golang) we set it to 1, so we never send the target region replica count as unset.


I think it'd be better if the builder file itself stopped parsing these target regions and adding these values, and instead we handle this inside of the

JenGoldstrich:
So to clarify, this is only for the global replica case, since setting the replica_count field is not allowed when setting target_regions

We actually send this currently for the build region if its not specified in the `replication_regions`, the plugin then adds it to the replication regions here https://github.com/hashicorp/packer-plugin-azure/blob/4d2a02095e1785c6df199f466ea8c70d42449f2e/builder/azure/arm/builder.go#L272-L279, so this is inconsistent based on when we specify the build location in that list or not, which is incorrect logic.

I think we should clean this up more, to just pass in the raw values the API expects, but I need to do more research and refactoring, so I am going to merge this for now to fix the bug

PR #443, opened 2024-09-24
fix error message

sujithq:
SigDestinationImageName and SigDestinationGalleryName were swapped in the error message. This was confusing.


JenGoldstrich (approved):
Hey @sujithq nice catch and thanks for this contribution

PR #445, opened 2024-10-03
Fixed some more typos.

lzandman:
I again fixed some typos. As in #391 I wasn't able to test `make generate` on my current system. I did not alter any files from `.web-docs`, so I hope the pipeline validation will succeed.


lzandman:
I guess I'm still doing something wrong...

builder/azure/chroot/shared_image_gallery_destination.go
```diff
@@ -21,7 +21,7 @@ type SharedImageGalleryDestination struct {
 
 	TargetRegions         []TargetRegion `mapstructure:"target_regions"`
 	ExcludeFromLatest     bool           `mapstructure:"exclude_from_latest"`
-	ExcludeFromLatestTypo bool           `mapstructure:"exlude_from_latest" undocumented:"true"`
+	ExcludeFromLatestTypo bool           `mapstructure:"exclude_from_latest" undocumented:"true"`
```
JenGoldstrich, line 24: Hey @lzandman this is a breaking change here, if you notice this field is called "ExcludeFromLatestTypo", and its labeled as undocumented, not really sure why this is here without doing some git spelunking, but can you revert this change?

builder/azure/chroot/shared_image_gallery_destination.hcl2spec.go
```diff
@@ -1,70 +0,0 @@
-// Code generated by "packer-sdc mapstructure-to-hcl2"; DO NOT EDIT.
```
JenGoldstrich, line 1: Was this file deleted intentionally? It's what is causing tests to fail.

JenGoldstrich (changes requested):
Thanks for this contribution!

Can you address the two comments I've made, afterwards I can run make generate for you if it still does not work on your system.

builder/azure/chroot/shared_image_gallery_destination.go
lzandman: I think I then have to revert all three `shared_image_gallery_destination*` files, right? They seem related.

builder/azure/chroot/shared_image_gallery_destination.hcl2spec.go
lzandman: I think I meant to remove it from the pull request. But deleted it entirely. Will try to revert.

lzandman: I've reverted the delete.

JenGoldstrich (approved):
I ran `go fmt ./...` to fix the last failing actions check on this PR and then pushed it to your fork, this looks good to me, thank you for this contribution @lzandman!

PR #450, opened 2024-10-29
Public IP Address Standard IP SKU Default

JenGoldstrich:
Follow up to https://github.com/hashicorp/packer-plugin-azure/pull/416

Now that Standard IP support has been out for a while and we've heard no reports of issues, its time to switch to defaulting to standard IPs before the March 2025 end of support for basic SKU IPs

### DIFF
--- a/.web-docs/components/builder/arm/README.md
+++ b/.web-docs/components/builder/arm/README.md
@@ -550,8 +550,8 @@ Providing `temp_resource_group_name` or `location` in combination with
   Requires enabling encryption at host in the Subscription read more [here](https://learn.microsoft.com/en-us/azure/virtual-machines/disks-enable-host-based-encryption-portal?tabs=azure-powershell)
 
 - `public_ip_sku` (string) - Specify the Public IP Address SKU for the public IP used to connect to the build Virtual machine.
-  Valid values are `Basic` and `Standard`. The default value is `Basic`.
-  On 31 March 2025 Azure will remove the ability to create `Basic` SKU public IPs, before then the plugin will be updated to create a Standard public IP by default
+  Valid values are `Basic` and `Standard`. The default value is `Standard`.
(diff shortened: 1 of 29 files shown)


lbajolet-hashicorp (approved):
LGTM! Left a tiny nit on the docs, but overall good to me

matt-richardson:
FYI: the change of defaults here caused us a bit of havoc... The instance metadata service behaviour changes, as [outlined here](https://learn.microsoft.com/en-us/azure/virtual-machines/instance-metadata-service?tabs=windows#sample-7-retrieve-public-ip-address).

JenGoldstrich:
Hey @matt-richardson I'm sorry to hear that! Azure is removing support for basic SKU IP Addresses (the norm before the launch of standard), which is forcing us to switch to this default sadly.  Do you think there is something we could have better documented on our end about this? Maybe we could include the note about breaking changes with IMDS.  We did release this in a minor instead of a patch, but I do want to make sure we document the risk of the havoc this can cause for other users upgrading

matt-richardson:
> Hey @matt-richardson I'm sorry to hear that! Azure is removing support for basic SKU IP Addresses (the norm before the launch of standard), which is forcing us to switch to this default sadly.  Do you think there is something we could have better documented on our end about this? Maybe we could include the note about breaking changes with IMDS.  We did release this in a minor instead of a patch, but I do want to make sure we document the risk of the havoc this can cause for other users upgrading

No stress - just wanted to make a note for future travellers. 

Completely understand that your hand was forced - either it breaks is now or it breaks us later.

TBH, I'm not sure you could have done anything different - the change and the impact are too far apart, so there's no place I would have looked to give me a heads up.

I think it's just the downside of us having a very loose version constraint.

_Maybe_ I might have seen a warning message logged saying the default had changed, but I'm not sure I would have noticed it.

Issue #451, opened 2024-10-30
Packer rejects valid OIDC token: client_jwt is missing the x5t header value


geekzter:
### Overview of the Issue

As Azure Pipelines, we are starting to issue Entra minted OIDC tokens. These tokens have a new format that is rejected by Packer while other tools can successfully use them.

### Reproduction Steps

Steps to reproduce this issue

Configure Azure Pipeline with AzureCLI task
Set `addSpnToEnvironment: true`
Run packer from within the task

`* client_jwt is missing the x5t header value, which is required for bearer JWT client authentication to Azure`

The same token can be used to authenticate to Azure successfully by Azure CLI and Terraform azure* providers.

It appears [this validation](https://github.com/hashicorp/packer-plugin-azure/blob/6c21022d98f477306b390613967ae63c35203daa/builder/azure/common/client/config.go#L218) should not be present.

### Plugin and Packer version

Packer v1.11.2  
packer-plugin-azure_v1.4.5_x5.0_linux_amd64

### Simplified Packer Buildfile

```hcl
variable "client_jwt" {
    type       = string
    default    = "${env("idToken")}"
    sensitive  = true
}

source "azure-arm" "image" {
    client_id  = "${var.client_id}"
    client_jwt = "${var.client_jwt}"
  ...
}
```

[Complete HCL](https://github.com/actions/runner-images/blob/main/images/windows/templates/windows-2022.pkr.hcl)
[Pipeline YAML](https://github.com/geekzter/azure-pipeline-agents/blob/f6b4b8d5a5042472ec7ed6b0f001ce91ef330f56/pipelines/build-image-isolated-template.yml#L349)

### Operating system and Environment details

Ubuntu 22.04 amd64

### Log Fragments and crash.log files

`* client_jwt is missing the x5t header value, which is required for bearer JWT client authentication to Azure`

JenGoldstrich:
Hey @geekzter thanks for letting us know about this issue, looking at the AzureRM TF provider I can definitely see that they don't enforce this check, so I think based on that and your use case this should be safe to remove.  I have merged a PR to remove this check, this issue was auto closed, I will close it shortly when a new release is created

JenGoldstrich:
Released in [v2.2.0]( https://github.com/hashicorp/packer-plugin-azure/releases/tag/v2.2.0), thanks again!

geekzter:
Thanks for the quick turnaround!

PR #452, opened 2024-11-04
Remove x5t Header Check

JenGoldstrich:
This check exists from the days of old, and does not seem relevant anymore.  When I search the Terraform AzureRM Provider for usages of the x5t header, the only references I see to it are in the now deprecated autorest library.  in the go-azure-sdk it is only referenced by the Microsoft Graph package, which this plugin does not use, so this should be safe to remove.

Closes https://github.com/hashicorp/packer-plugin-azure/issues/451

### DIFF
--- a/builder/azure/common/client/config.go
+++ b/builder/azure/common/client/config.go
@@ -211,15 +211,10 @@ func (c Config) Validate(errs *packersdk.MultiError) {
 		c.ClientJWT != "" {
 		p := jwt.Parser{}
 		claims := jwt.StandardClaims{}
-		token, _, err := p.ParseUnverified(c.ClientJWT, &claims)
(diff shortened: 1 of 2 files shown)


lbajolet-hashicorp (approved):
Looking at the JWT spec quickly, it seems that you need to have a x5t or a kid for the JWT to be valid.
That said, the validation can be left as a responsibility to the Azure APIs, since they'll have to make sure the JWT is valid before accepting it, so I trust they do a better job at sticking to the specification than us.

In the end, LGTM over here, this'll likely accept more tokens, I'm not knowledgeable enough about JWTs to ensure that we don't too broadly accept those tokens, but I trust Azure will be the ones doing the validation on the token upstream so we should be safe.

Issue #431, opened 2024-07-10
Add a property for the full gallery ID to the shared_image_gallery block


danbarr:
#### Community Note

Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request.
Please do not leave "+1" or "me too" comments, they generate extra noise for issue followers and do not help prioritize the request.
If you are interested in working on this issue or have submitted a pull request, please leave a comment.

#### Description

Currently, the `shared_image_gallery` block has individual properties for the elements of an Azure Compute Gallery's ID: `subscription`, `resource_group`, `gallery_name`, `image_name`, `image_version`.

But when referencing a source/parent image from HCP Packer, the `hcp-packer-artifact` data source's external_identifier returns the full ID string, `/subscriptions/<id>/resourceGroups/<name>/providers/Microsoft.Compute/galleries/<name>/images/<name>/versions/<version>`

This means we have to do some nasty & fragile string parsing to pull out the individual components, or use the build labels from the HCP Packer artifact like this:

```
  # Source image
  shared_image_gallery {
    subscription   = var.az_subscription_id
    resource_group = data.hcp-packer-artifact.example.labels.sig_resource_group
    gallery_name   = data.hcp-packer-artifact.example.labels.sig_name
    image_name     = data.hcp-packer-artifact.example.labels.sig_image_name
    image_version  = data.hcp-packer-artifact.example.labels.sig_image_version
  }
```

Propose the addition of a `shared_image_gallery_id` attribute which can be used in lieu of the five individual attributes, similar to the existing `community_gallery_image_id` and `direct_shared_gallery_image_id` attributes.

#### Use Case(s)

Integrating with HCP Packer to resolve a source image that is located in a Compute Gallery.

#### Potential configuration

```
  # Source image
  shared_image_gallery {
    shared_image_gallery_id = data.hcp-packer-artifact.example.external_identifier
  }
```

PR #482, opened 2025-03-28
Allow parent SIG images to referenced by their ID

JenGoldstrich:
Allow SIG base image to be specified by resource ID, this is specifically useful when using HCP Packer with its ancestry feature.

Prior to this PR you had to set a bunch of fields inside of the shared_image_gallery block to set the base image.  When using HCP Packer this looked like
```
 shared_image_gallery {
    subscription   = var.az_subscription_id
    resource_group = data.hcp-packer-artifact.example.labels.sig_resource_group
    gallery_name   = data.hcp-packer-artifact.example.labels.sig_name
    image_name     = data.hcp-packer-artifact.example.labels.sig_image_name
    image_version  = data.hcp-packer-artifact.example.labels.sig_image_version
  }
```

After this PR, you can 
```
 shared_image_gallery {
    id = data.hcp-packer-artifact.parent.external_identifier 
  }
 ```

This is an ask from last year that has a significant amount of thumbs up, and requests from users internally.

closes https://github.com/hashicorp/packer-plugin-azure/issues/431

### DIFF
--- a/.web-docs/components/builder/arm/README.md
+++ b/.web-docs/components/builder/arm/README.md
@@ -634,6 +634,8 @@ The shared_image_gallery block is available for building a new image from a priv
 
 <!-- Code generated from the comments of the SharedImageGallery struct in builder/azure/arm/config.go; DO NOT EDIT MANUALLY -->
 
+- `id` (string) - ID
+
 - `subscription` (string) - Subscription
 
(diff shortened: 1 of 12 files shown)


tanmay-hc (approved):
Changes looks good to me!

PR #488, opened 2025-04-03
Add temp resource names to Generated Data

### LINKED ISSUE #369: Create new output vars
#### Description

Add/create output variables that contain the name of the resources packer spun up. If packer spun up a VM/NIC/HDD then it'd look something like the below still following the "Randomized" logic already in place.

e.g.

```bash
build.vm_name=pkrvm*
build.nic_name=pkrni*
build.osdisk_name=pkros*
etc
etc
```

#### Use Case(s)

The use case I've had in mind is if packer fails or the pipeline run is canceled it leaves behind resources it provisioned.

If I knew or could know the name of the resources it spun up I could write a simple shell script to be executed on packer failure to ensure the resources it spun up are indeed gone.

#### Potential configuration

NA

#### Potential References
NA

JenGoldstrich:
Add temp resource names to ARM builder generated data.

Full list:

- `TempComputeName` - Virtual machine name
- `TempNicName` - Network interface name
- `TempOSDiskName` - OS Disk Name
- `TempDataDiskName` - Data Disk Name
- `TempDeploymentName` - ARM Deployment name
- `TempVirtualNetworkName` - Virtual Network name
- `TempKeyVaultName` - Key vault name
- `TempResourceGroupName` - Resource Group Name, will be unset if a build resource group name is provided.
- `TempNsgName` - Network security group name
- `TempSubnetName` - Subnet name
- `TempPublicIPAddressName` - Public IP Address name

Closes #369

### DIFF
--- a/.web-docs/components/builder/arm/README.md
+++ b/.web-docs/components/builder/arm/README.md
@@ -742,6 +742,22 @@ shared images the resulting name will point to the actual source used to create 
 
 - `TenantID` - The ID of the Azure Tenant where the build takes place.
 
+All of the following variables are temporary resource names that the plugin uses to build resources that are generally deleted at the end of a build.
+
+The presence of these variables does not mean these resources are created, they are just the names that are used for any temporary resources the builder creates. They are all randomly specified, but most can be overridden by builder fields.
+
+- `TempComputeName` - Virtual machine name
+- `TempNicName` - Network interface name
+- `TempOSDiskName` - OS Disk Name
+- `TempDataDiskName` - Data Disk Name
+- `TempDeploymentName` - ARM Deployment name
+- `TempVirtualNetworkName` - Virtual Network name
+- `TempKeyVaultName` - Key vault name
+- `TempResourceGroupName` - Resource Group Name, will be unset if a build resource group name is provided.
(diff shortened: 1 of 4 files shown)


jsloan117:
@JenGoldstrich  wow that was fast and thank you so much!!!

For clarity - if for example, I define in source.pkr.hcl `build_key_vault_name = "my_key_vault_name"` would  `${build.TempKeyVaultName}` contain this value? Or does that variable exist only when packer automatically calculates this value on the users behalf?

I ask because in my log, `TempKeyVaultName` doesn't exist, but the below does.

```
2025/04/02 18:37:32 packer-plugin-azure_v2.3.2_x5.0_linux_amd64 plugin: 2025/04/02 18:37:32 BuildKeyVaultName=my_key_vault_name
```

jsloan117:
Excellent information, and I appreciate the extra context!

As the linked issue stated, we didn't want to be forced to configure the VM,NIC,DISK settings to ensure we knew their values. In the event resources weren't deleted we wanted to setup automation to do so. Which this should allow me to do just that.

This makes sense to me and LGTM!

galapas1 (approved):

PR #490, opened 2025-04-14
Disable DTL acceptance tests

JenGoldstrich:
Dev Test Labs tests recently started failing after the April 1st removal of support for basic IP skus.  This is because Azure subnets without network security groups will now silently fail for labs.  Labs have very low usage in our internal telemetry (less than a percent, and we suspect a lot of this usage is from our internal runs of the acceptance tests, and not user usage).  There are very few user issues reported on DTL historically, and there is not much activity or blogs from Azure around this feature in a very long time. 

To get DTL tests to pass you must create a network security group that allows ingress from the runner's IP, or with very permissive rules, basic IPs behaved this way by default but this always had lax security implications.  Since both of these solutions have their costs and trade offs, and DTL is very under used, I think it is best to disable these two tests and focus developer resources on the builders with more user activity (ARM, chroot)

When I started working on the plugin a few years ago these tests were not being regularly run, so I do not feel like this

### DIFF
--- a/builder/azure/dtl/builder_acc_test.go
+++ b/builder/azure/dtl/builder_acc_test.go
@@ -30,7 +30,12 @@ import (
 	"github.com/hashicorp/packer-plugin-sdk/acctest"
 )
 
+// DTL Tests were paused in April 2025 after the deprecation of Basic IP SKUs
+// This requires creating a network security group with either overly permissive ingress, or the expected clients IP
+// The complexity of maintaining this in CI is not worth the time currently when there are higher priority problems on the ARM builder, and DTL is very inactive
+
 func TestDTLBuilderAcc_ManagedDisk_Windows(t *testing.T) {
+	t.Skip()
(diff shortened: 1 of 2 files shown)


builder/azure/dtl/builder_acc_test.go
```diff
@@ -30,7 +30,12 @@ import (
 	"github.com/hashicorp/packer-plugin-sdk/acctest"
 )
 
+// DTL Tests weer paused in April 2025 after the deprecation of Basic IP SKUs
```
lbajolet-hashicorp, line 33: Mini typo

```suggestion
// DTL Tests were paused in April 2025 after the deprecation of Basic IP SKUs
```

lbajolet-hashicorp (approved):
LGTM, left a mini nit on a comment, but all good besides that

lbajolet-hashicorp (approved):

PR #493, opened 2025-04-16
Update Golang-JWT to v5.2.2

JenGoldstrich:


builder/azure/common/client/config.go
```diff
@@ -209,12 +206,6 @@ func (c Config) Validate(errs *packersdk.MultiError) {
 		c.ClientSecret == "" &&
 		c.ClientCertPath == "" &&
 		c.ClientJWT != "" {
-		p := jwt.Parser{}
```
JenGoldstrich, line 212: Standard Claims are no longer supported, we could use map claims, but honestly this code feels extraneous

galapas1 (approved):

PR #494, opened 2025-04-17
Enforce /x/net 0.38

JenGoldstrich:


lbajolet-hashicorp (approved):
LGTM, but may I ask why the update? I've seen a CVE report recently that targetted versions prior to 0.36.0 of the module, is there something else I missed?

JenGoldstrich:
There's a new one for 0.38.0 https://github.com/hashicorp/packer-plugin-azure/security/dependabot/35

PR #508, opened 2025-07-08
BUG: Fetching the Blob Endpoint for Deletion from Account

### LINKED ISSUE #502: VHD builds fail in USGovCloud due to hardcoded blob endpoint
#### Overview of the Issue
We currently are attempting to build VHDs inside of Azure US gov cloud however VHD builds fail to complete correctly. Setting PACKER_LOG=1 reveals that the packer plugin is trying to delete vhds from **blob.core.windows.net** instead of **blob.core.usgovcloudapi.net** despite setting cloud_environment_name to AzureUSGovernmentCloud. After poking around, I found that the [plugin is hardcoding the azure blob endpoint in the VHD cleanup](https://github.com/hashicorp/packer-plugin-azure/blob/main/builder/azure/arm/azure_client.go#L238C3-L238C128). This means the packer build of VHDs is only supported for azure public cloud. We had to fork the code to point it to govcloud's blob endpoint. Ideally it should respect the cloud_environment_name value and set the blob endpoint [similar to how it's being handled here](https://github.com/hashicorp/packer-plugin-azure/blob/019929fc2e79f0c91e98161038c825a91a002cd8/builder/azure/arm/builder.go#L200).

#### Reproduction Steps

Build an os image off vhds in usgovcloud.

### Plugin and Packer version

Plugin: 2.3.3
Packer: 1.13.1

tanmay-hc:
The code uses a separate SDK client for the VHD deletion, where the blob endpoint is hardcoded as `.blob.core.windows.net`. This causes errors while deleting VHDs for a different cloud env.

This PR updates the blob endpoint with the one we get from the storage account properties, which is already being used to create a new artifact.

Closes #502

### DIFF
--- a/builder/azure/arm/builder.go
+++ b/builder/azure/arm/builder.go
@@ -126,11 +126,10 @@ func (b *Builder) Run(ctx context.Context, ui packersdk.Ui, hook packersdk.Hook)
 		b.config.PollingDurationTimeout,
 		authOptions,
 	)
-
-	ui.Message("ARM Client successfully created")
 	if err != nil {
 		return nil, err
 	}
+	ui.Message("ARM Client successfully created")
 
 	resolver := newResourceResolver(azureClient)
 	if err := resolver.Resolve(&b.config); err != nil {
@@ -197,7 +196,12 @@ func (b *Builder) Run(ctx context.Context, ui packersdk.Ui, hook packersdk.Hook)
 		if err != nil {
 			return nil, err
 		}
-		b.config.storageAccountBlobEndpoint = *account.Properties.PrimaryEndpoints.Blob
+
+		blobPrimaryEndpoint := *account.Properties.PrimaryEndpoints.Blob
+		b.config.storageAccountBlobEndpoint = blobPrimaryEndpoint
+		// Remove trailing slash
+		azureClient.GiovanniBlobClient.Client.BaseUri = blobPrimaryEndpoint[:len(blobPrimaryEndpoint)-1]
+
 		if !equalLocation(account.Location, b.config.Location) {
 			return nil, fmt.Errorf("The storage account is located in %s, but the build will take place in %s. The locations must be identical", account.Location, b.config.Location)
 		}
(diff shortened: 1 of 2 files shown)


builder/azure/arm/builder.go
```diff
@@ -197,7 +196,12 @@ func (b *Builder) Run(ctx context.Context, ui packersdk.Ui, hook packersdk.Hook)
 		if err != nil {
 			return nil, err
 		}
-		b.config.storageAccountBlobEndpoint = *account.Properties.PrimaryEndpoints.Blob
+
+		blobPrimaryEndpoint := *account.Properties.PrimaryEndpoints.Blob
+		b.config.storageAccountBlobEndpoint = blobPrimaryEndpoint
+		// Remove trailing slash
+		azureClient.GiovanniBlobClient.Client.BaseUri = blobPrimaryEndpoint[:len(blobPrimaryEndpoint)-1]
```
JenGoldstrich, line 203: This change looks good to me, but it feels a bit odd to leave behind the now code in the arm/azure_client file, where we set https://github.com/hashicorp/packer-plugin-azure/blob/a8d4c7193fad581b1640d5e6dd606524d6c2e726/builder/azure/arm/azure_client.go#L238-L241 here.

IMO we should only set the BaseUri once for this client, rather than overriding it with the valid value after the fact, maybe we could set the blobPrimaryEndpoint before we call the ARM Client, and then use that blobPrimaryEndpoint in the azure_client file, and remove line 203 here

What do you think?

tanmay-hc: @JenGoldstrich That is what I was trying to do initially, however, the client creation is required first in order to fetch these properties.

In order to move this call to the client initiation, we would have to update the code to pass the resource group info somehow to the function.
Or, if we don't want to override, we can just use the following as the placeholder.
```
blobClient, err := giovanniBlobStorageSDK.NewWithBaseUri("") 
```

JenGoldstrich: I see, that makes sense, do you think you could just add a comment to the azure client to indicate that this value is required, but will be overriden?

JenGoldstrich (approved):
Approving this, please consider adding the comment though, but its just a suggestion and not critical, nice work here friend!

builder/azure/arm/builder.go
tanmay-hc: I've added a comment for the same. Please have a look.

PR #509, opened 2025-07-09
BUG: Prevents snapshot overwrite conflicts in managed images

### LINKED ISSUE #459: ARM Snapshot overwrite fails when using '-force'
#### Overview of the Issue

As described here:
[https://developer.hashicorp.com/packer/docs/commands/build#force](https://urldefense.com/v3/__https://developer.hashicorp.com/packer/docs/commands/build*force__;Iw!!Pe07N362zA!296vurIKIKgNj-mqjmfkOG2XcHfmsjNDObtoLLgeaEypwN1WMt8YtRTBII51axsThL_Bad-tqAwdZeLZfDU$)

The '-force' flag should remove/overwrite the artifacts created by previous builds. When applying it for azure-arm packer build, it is indeed able to overwrite an existing image name, but if ‘managed_image_os_disk_snapshot_name’ & ‘managed_image_data_disk_snapshot_prefix’ are applied, they are not being overwritten and the build fails:

2024-12-30T17:00:29Z: Build 'azure-arm.linux-component' errored after 4 minutes 33 seconds: performing CreateOrUpdate: unexpected status 400 (400 Bad Request) with error: BadRequest: Changing property 'sourceResourceId' is not allowed for existing disk 'testimagename'.

#### Reproduction Steps

Execute packer build with azure-arm provisioner, while applying 'managed_image_os_disk_snapshot_name' / ‘managed_image_data_disk_snapshot_prefix’, then rerun it using the same values with the '-force' flag.

### Plugin and Packer version

Packer v1.10.0
github.com/hashicorp/azure v2.2.0

https://gist.github.com/pelegor/4461ea8125f98ba0e122fc5cc805d8d7

Image specs:
["RedHat", "RHEL", "92-gen2", "latest"]

https://gist.github.com/pelegor/3cf659e27d277c9211aa5120b1a9449f

tanmay-hc:
Adds checks for existing OS and data disk snapshots to avoid overwrite errors when creating managed images. Deletes existing snapshots if the `-force` flag is set, or fails with a clear error otherwise. Improves reliability when reusing image names.

Closes #459

### DIFF
--- a/builder/azure/arm/builder.go
+++ b/builder/azure/arm/builder.go
@@ -10,11 +10,13 @@ import (
 	"fmt"
 	"os"
 	"runtime"
+	"strconv"
 	"strings"
 
 	"github.com/hashicorp/packer-plugin-azure/builder/azure/common/log"
 
 	"github.com/hashicorp/go-azure-sdk/resource-manager/compute/2022-03-01/images"
+	"github.com/hashicorp/go-azure-sdk/resource-manager/compute/2022-03-02/snapshots"
 	"github.com/hashicorp/go-azure-sdk/resource-manager/compute/2023-07-03/galleryimages"
 	"github.com/hashicorp/go-azure-sdk/resource-manager/compute/2023-07-03/galleryimageversions"
 	"github.com/hashicorp/go-azure-sdk/resource-manager/network/2023-09-01/publicipaddresses"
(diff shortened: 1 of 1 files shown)


builder/azure/arm/builder.go
```diff
@@ -178,6 +179,36 @@ func (b *Builder) Run(ctx context.Context, ui packersdk.Ui, hook packersdk.Hook)
...
+				if err != nil {
+					return nil, fmt.Errorf("failed to delete the managed image snapshot named %s : %s", b.config.ManagedImageOSDiskSnapshotName, azureClient.LastError.Error())
+				}
+			} else {
+				return nil, fmt.Errorf("the managed image os disk snapshot named %s already exists in the resource group %s, use a different manage image name or use the -force option to automatically delete it.", b.config.ManagedImageOSDiskSnapshotName, b.config.ManagedImageResourceGroupName)
+			}
+		}
+
+		// If a managed image Data disk snapshot already exists it cannot be overwritten.
+		snapshotId = snapshots.NewSnapshotID(b.config.ClientConfig.SubscriptionID, b.config.ManagedImageResourceGroupName, b.config.ManagedImageDataDiskSnapshotPrefix)
```
JenGoldstrich, line 199: I know you haven't asked me for review yet so sorry if this is premature but I was taking a look and noticed how you're checking for the datadisk snapshot, this may be a bit tricky but this isn't the name of datadisk snapshots

https://github.com/hashicorp/packer-plugin-azure/blob/a8d4c7193fad581b1640d5e6dd606524d6c2e726/builder/azure/arm/step_snapshot_data_disks.go#L86-L87

Data disk snapshots will have the prefix and then they'll be appended by a number,  so if the prefix is "jenna", the data disks would be "jenna1", "jenna2", ..."jenna10" depending on how many data disks there.

 I think this may be require a bit of thinking on how best to search for those images, but I think this won't work as Azure will search for the ID specifically, and not for things that are similar to the ID.

tanmay-hc: Thank you for pointing this out. I have made a fix for this. Please have a look 😃

JenGoldstrich (approved):
LGTM, nice work here Tanmay!

PR #511, opened 2025-07-15
Updated code that validates managed_image_os_disk_snapshot_name and managed_image_data_disk_snapshot_prefix

VladimirKhvostov:
### Description
MyImage2025.01.01-snapshot is a perfectly valid snapshot name, but original code does not accept it. Original code does not allow to use dots in the snapshot name.

No


JenGoldstrich:
Hey @VladimirKhvostov this change seems good to me, thank you for the unit test changes as well.

Can you please sign the CLA mentioned in the previous comment? Once that's done I will validate it quickly and then we should be good to merge this

PR #516, opened 2025-07-21
FEAT: Added KeyVaultSecret Datasource

### LINKED ISSUE #71: Add support for specifying key to fetch from Azure Key Vault
_This issue was originally opened by @ricardclau as hashicorp/packer#9404. It was migrated here as a result of the [Packer plugin split](https://github.com/hashicorp/packer/issues/8610#issuecomment-770034737). The original body of the issue is below._

<hr>

#### Feature Description

Very similar to the relatively recent https://github.com/hashicorp/packer/issues/9086 but in this case using Azure KV service

#### Use Case(s)

Same case as the mentioned feature, ability to use passwords or other secrets stored there

Not sure if there is enough interest for such feature but happy to work on this if nobody can step in and other fellow members are interested. The terraform provider can already interact with this so probably most of the heavy work is done there and it would be adapting it to the packer codebase

tanmay-hc:
This pull request introduces a new Key Vault data source for the Azure Packer plugin, enabling users to retrieve secrets from Azure Key Vault.

- `vault_name` [required]
- `secret_name` [required]
- `version`

NOTE: Unlike the rest of the plugin, which uses the Hashicorp's `go-azure-sdk`, this change uses the Microsoft "azure-sdk-for-go" for the acceptance tests, this will be changed to use go-azure-sdk throughout in a later change. The importing of azure-sdk-for-go is only used in acceptance tests, and does not impact the code

Closes #71

### DIFF
--- a/.github/workflows/acceptance-tests.yaml
+++ b/.github/workflows/acceptance-tests.yaml
@@ -6,7 +6,6 @@ on:
   push:
     branches:
     - main
-    - keyvault-datasource
 
 permissions:
   contents: read


datasource/keyvaultsecret/data_acc_test.go
```diff
@@ -0,0 +1,223 @@
+// // Copyright (c) HashiCorp, Inc.
+// // SPDX-License-Identifier: MPL-2.0
+
+package keyvaultsecret
```
JenGoldstrich, line 4: This is a very well written test, nice work Tanmay!

datasource/keyvaultsecret/data_acc_test.go
```diff
@@ -0,0 +1,223 @@
...
+						return nil
+					}
+					return tc.secret.Create()
+				},
+				Teardown: func() error {
+					if tc.skipCreate {
+						return nil
+					}
+					return tc.secret.Delete()
+				},
```
JenGoldstrich, line 102: Smart!

datasource/keyvaultsecret/data.go
```diff
@@ -0,0 +1,142 @@
...
+	}
+
+	if d.config.TenantID == "" {
+		d.config.TenantID = os.Getenv("AZURE_TENANT_ID")
+	}
+	if d.config.ClientID == "" {
+		d.config.ClientID = os.Getenv("AZURE_CLIENT_ID")
+	}
+	if d.config.ClientSecret == "" {
+		d.config.ClientSecret = os.Getenv("AZURE_CLIENT_SECRET")
```
JenGoldstrich, line 91: I assume that AZURE_TENANT_ID/etc are environment variables that the Azure SDK uses, since they aren't used elsewhere in the plugin this behavior may be confusing, we should document it clearly, but preferably we wouldn't use one type of authentication setup in one Packer Azure resource that we aren't using in the other builders

datasource/keyvaultsecret/data.go
```diff
@@ -0,0 +1,142 @@
...
+	}
+
+	if errs != nil && len(errs.Errors) > 0 {
+		return errs
+	}
+	return nil
+}
+
+func (d *Datasource) Execute() (cty.Value, error) {
```
JenGoldstrich, line 100: Super small nit, but future developers may find it confusing as to why we use two different SDKs, I understand we had to for this use case, but I think it'd be good to document that here in a comment

datasource/keyvaultsecret/data.go
```diff
@@ -0,0 +1,142 @@
...
+
+	var cred azcore.TokenCredential
+	var err error
+
+	if d.config.TenantID != "" && d.config.ClientID == "" && d.config.ClientSecret == "" {
+		log.Printf("Using ClientSecretCredential for vault %q", d.config.VaultName)
+		cred, err = azidentity.NewClientSecretCredential(d.config.TenantID, d.config.ClientID, d.config.ClientSecret, nil)
+	} else {
+		log.Printf("Using DefaultAzureCredential for vault %q", d.config.VaultName)
+		cred, err = azidentity.NewDefaultAzureCredential(nil)
```
JenGoldstrich, line 110: I think that calling the sdk's auth library like this won't support all of the same authentication methods that the plugin builder currently supports.  For example OIDC authentication won't work without the ability to define the expected oidc_request_url field I think.

I think you should more closely model the authentication here off how the builders handle it, even using the same fields.  While it may seem silly that having for example the use_azure_cli_auth field twice in one template, its possible users will only be using this datasource without the builder.  The docs for the main plugin https://developer.hashicorp.com/packer/integrations/hashicorp/azure#authentication list authentication fields that are implied to be plugin wide

The Packer AWS plugin has a good example here where common auth object with the same fields is shared between builders and datasources https://github.com/hashicorp/packer-plugin-amazon/blob/main/common/access_config.go.  I think this is the kind of experience we should aim for.  I know it doesn't mesh very well with the Azure SDK neccesarily, and if there's anything we can't do without a lot of overhaul, I think its worth just stating that some authentication type is not supported on this datasource, but we should use the same fields as the builder IMO for the authentication mechanics we do explicitly support

What do you think?

JenGoldstrich (commented):
Nice work Tanmay, this is looking really good.  I think we need to re-think some of the authentication logic and bring it more in line with the how the builders work, as I think this will be a better user experience, please let me know what you think!

datasource/keyvaultsecret/test-fixtures/template.pkr.hcl
```diff
@@ -0,0 +1,53 @@
...
+  type = string
+}
+
+data "azure-keyvaultsecret" "test" {
+  vault_name = var.vault_name
+  secret_name   = var.secret_name
+
+  client_id     = var.client_id
+  client_secret = var.client_secret
+  tenant_id     = var.tenant_id
```
JenGoldstrich, line 30: I had to add subscription_id here after the changes to use the existing authorizer

datasource/keyvaultsecret/data_acc_test.go
```diff
@@ -0,0 +1,220 @@
...
+	"log"
+	"os"
+	"os/exec"
+	"regexp"
+	"testing"
+	"time"
+
+	"github.com/Azure/azure-sdk-for-go/sdk/azcore"
+	"github.com/Azure/azure-sdk-for-go/sdk/azidentity"
+	"github.com/Azure/azure-sdk-for-go/sdk/keyvault/azsecrets"
```
JenGoldstrich, line 20: Do we still need to use the azure-sdk-for-go here? I assume you just haven't updated this acceptance test since migrating to the new strategy for solving this

terraform/main.tf
```diff
@@ -52,6 +52,25 @@ resource "azurerm_shared_image" "linux-sig" {
   }
 }
 
+resource "azurerm_key_vault" "vault" {
+  name                        = "packer-acctest-vault"
```
JenGoldstrich, line 56: Right now if two developers are working on the plugin, or if CI is running when a developer is working, this can cause conflict, you should use the resource prefix for resources that have unique name across subscription, which I think key vaults are at least subscription wide for unique names, since we ran into that error before

```suggestion
  name                        = "${var.resource_prefix}-pkr-test-vault"
```

Issue #529, opened 2025-08-19
shared_image_gallery.id does not match expected format error when running packer validate


jsloan117:
When filing a bug, please include the following headings.

Please delete the example text in this template before submitting.

#### Overview of the Issue

Template validation fails using the new HCP Packer `shared_image_gallery.id` syntax on a child image.

```
shared_image_gallery.id does not match expected format of '/subscriptions/(subscriptionid)/resourceGroups/(rg-name)/providers/Microsoft.Compute/galleries/(gallery-name)/images/image-name/versions/(version)
```

#### Reproduction Steps

packer validate my_template.pkr.hcl

### Plugin and Packer version

Packer v1.14.1
packer-plugin-azure_v2.4.0_x5.0_linux_amd64

### Simplified Packer Buildfile

Our template is failing, and I tried using the new testdata file [here](https://github.com/hashicorp/packer-plugin-azure/blob/main/builder/azure/arm/testdata/example_hcp_ancestry.pkr.hcl).

### Operating system and Environment details

Ubuntu 22.04 Jammy
x64
Executed via Azure DevOps pipelines and tested locally via Windows PowerShell.

### Log Fragments and crash.log files

https://gist.github.com/jsloan117/36fa792adc9517fb6d2e692be0ad9821

jsloan117:
If I pass `id = 'hardcoded_id`, it works fine. I'm starting to think the automatically calculated value is wrong. Maybe it's passing in some ManagedImageId despite not using a Managed Image. In both the parent & child packer templates, we're using shared_image_gallery as a source.

JenGoldstrich:
Does your parent build build its final image into a managed image though? You mention using SIG as a source and say you don't use managed image, but in your parent template is there a reference to `managed_image_name` in your parent template?  This is the only situation where a managed image would be set as the ID here I believe.

Can you include the external identifier of the Packer build of the parent build, this can be found in the artifact card on the HCP Packer UI, can you also include the parent template?

<img width="621" height="161" alt="Image" src="https://github.com/user-attachments/assets/38edf717-0e10-4043-beae-9ed17a4a136e" />

jsloan117:
We don't have `managed_image_name` in any templates. I'll have to open a ticket with HCP I can't provide those details here.

JenGoldstrich:
[...]

PR #531, opened 2025-08-28
Move SIG Regex check to builder to fix validation failures

JenGoldstrich:
When running `packer validate` the HCP Packer Datasource does not make API requests, this is to avoid generating billable events for users who are just validating templates.  This causes a problem when doing regex validation on the source image SIG ID, this lead to @jsloan117 reporting issues, I was unable to reproduce at first, but then Jessica pointed out that all of my testing was using `packer build`, while she was using `packer validate`, as soon as I tried packer validate the issue immediately was apparent, and also why when she hardcodes the ID this issue goes away.

To fix this issue, I moved this regex comparison to outside of the config function that the validate template command calls.  So that the SIG ID is not validated when it might not be populated.

Closes https://github.com/hashicorp/packer-plugin-azure/issues/529

Before this change I see the following error:

```
packer validate builder/azure/arm/testdata/child_from_specialized_parent.pkr.hcl
Error: 1 error(s) occurred:

* shared_image_gallery.id does not match expected format of '/subscriptions/(subscriptionid)/resourceGroups/(rg-name)/providers/Microsoft.Compute/galleries/(gallery-name)/images/image-name/versions/(version)

  on builder/azure/arm/testdata/child_from_specialized_parent.pkr.hcl line 40:
  (source code not available)
```

After this change, 
```
packer validate builder/azure/arm/testdata/child_from_specialized_parent.pkr.hcl
The configuration is valid.
```

### DIFF
--- a/builder/azure/arm/builder.go
+++ b/builder/azure/arm/builder.go
@@ -9,6 +9,7 @@ import (
 	"errors"
 	"fmt"
 	"os"
+	"regexp"
 	"runtime"
 	"strconv"
 	"strings"
@@ -372,6 +373,9 @@ func (b *Builder) Run(ctx context.Context, ui packersdk.Ui, hook packersdk.Hook)
 		}
 	}
 	if b.config.SharedGallery.ID != "" {
+		if !b.IsSigIDValid() {
+			return nil, errors.New(fmt.Sprintf("shared_image_gallery.id (%s) does not match expected format of '/subscriptions/(subscriptionid)/resourceGroups/(rg-name)/providers/Microsoft.Compute/galleries/(gallery-name)/images/image-name/versions/(version)", b.config.SharedGallery.ID))
+		}
 		sigID := b.config.getSharedImageGalleryObjectFromId()
 		if sigID == nil {
 			// The SIG ID should have already been validated at this point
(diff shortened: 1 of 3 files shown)


tanmay-hc (approved):
LGTM!

JenGoldstrich:
@jsloan117 Glad this fixed your issue, do you mean why can't you say "give me v2" from the HCP Packer datasource?

Generally we encourage users to use channels, this way users can avoid having to change their templates when a version is revoked.  This allows HCP Packer administrators to control which images are used by different teams, without them having to have edit permissions to each team's templates.  Often those who manage the image estate, may not have permissions over the Packer templates.  If you manually target v2, and v2 is revoked, then your template will start failing, but if you target the "prod" channel, and the image management team promotes the newest valid image always to prod, the team running `packer builds` don't have to worry about what version of the artifact they're using, they just get the latest approved image.

JenGoldstrich:
I'm merging this PR, but feel free to reach out to me, or open a new issue with any further issues or questions!

Thanks again for your smart validation!

JenGoldstrich:
Oh sorry one question I missed, yes the SIG ID change from my previous PR is added to the now moved error message

JenGoldstrich:
Ah I got you now, yes so the image version is meant to be an incremental version, using the image version would be helpful, but could lead to weird race conditions, versus a system controlled version that we can handle concurrent requests on.  

Many image types supported by HCP Packer do not have their own built in versioning scheme, and our versioning logic is meant to be consistent regardless of the type of image/platform you build on.

PR #534, opened 2025-08-28
VHD Migrations Fixes - VHD Acceptance Tests, and Disk Revoke Access on Failures

JenGoldstrich:
This PR contains 2 main changes

1. After the VHD unmanaged disk removals, acceptance tests started failing, this is due to the capture container name expected to be created before a build is started now, to fix this I added the capture container we'll use to the acceptance testing Terraform
2. I also made a change to fix an issue caused by DiskRevokeAccess only being called after a copy is successful.  If the copy fails (such as due to the above error where a user doesn't know they need to create a container now), we still need to Revoke the DIsk Access, otherwise deletion of the build disk will fail.  To fix this I've moved this into the Cleanup of the CaptureImage step

### DIFF
--- a/.web-docs/components/builder/arm/README.md
+++ b/.web-docs/components/builder/arm/README.md
@@ -147,7 +147,7 @@ Providing `temp_resource_group_name` or `location` in combination with
 
 - `capture_name_prefix` (string) - VHD prefix.
 
-- `capture_container_name` (string) - Destination container name.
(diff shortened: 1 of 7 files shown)


tanmay-hc (approved):
Hi Jenna,
Thanks for fixing this. The changes looks good to me.
Also tried the acceptance tests and they are working now.

PR #536, opened 2025-09-01
version: cut release 2.5.0

tanmay-hc:


CHANGELOG.md
```diff
@@ -3,6 +3,24 @@
 Please refer to [releases](https://github.com/hashicorp/packer-plugin-azure/releases) for the latest CHANGELOG information.
 
 ---
+## 2.5.0 (September 3, 2025)
+
+## What's Changed
+
```
JenGoldstrich, line 9:
```suggestion
### Breaking Changes
* VHDs are no longer built using unmanaged disks, this change was made because of the following deprecation notice from Microsoft https://learn.microsoft.com/en-us/azure/virtual-machines/unmanaged-disks-deprecation.  Users must now create the capture container in their storage account at the root of the container.  The plugin no longer relies on Azure system created capture containers.
```

Maybe something like this?

kp2099 (approved):

PR #552, opened 2025-12-09
Fix: Update number of allowed resource tags

obabec:
Azure nowadays supports 50 tags per resource -- https://learn.microsoft.com/en-us/azure/azure-resource-manager/management/tag-resources

This PR updates the validation to allow 50 tags.


tanmay-hc (dismissed):

tanmay-hc:
@obabec I plan to merge and do a minor release in around 10 days from now, before the holidays.

obabec:
@tanmay-hc That should be ok. Is there something else that I need to do? Looks like the tests were not triggered for my last commit...

obabec:
@tanmay-hc done :)

tanmay-hc (approved):

PR #579, opened 2026-02-13
relax parameter checks (skip_create_image true, storage account location, rg existence, etc.)

arsdragonfly:
### Description
Relax parameter checks so that

- skip_create_image can work without any destination parameter specified
- storage account for VHD can be in different region (this is a leftover check from pre-managed-disk era)
- managed image can be created into the temporary RG
- relax managed disk SAS expiry to avoid VHD copy timeout (especially considering how smaller Azure OS disks have horrendous throughput)

### DIFF
--- a/builder/azure/arm/config.go
+++ b/builder/azure/arm/config.go
@@ -1180,6 +1180,10 @@ func assertRequiredParametersSet(c *Config, errs *packersdk.MultiError) {
 		errs = packersdk.MultiErrorAppend(errs, fmt.Errorf("A capture_container_name, managed_image_name and managed_image_resource_group_name, or shared_image_gallery_destination must be specified"))
 	}
 
+	if (c.ManagedImageName == "") != (c.ManagedImageResourceGroupName == "") {
+		errs = packersdk.MultiErrorAppend(errs, fmt.Errorf("managed_image_name and managed_image_resource_group_name must be specified together"))
+	}
+
 	if c.CaptureContainerName != "" {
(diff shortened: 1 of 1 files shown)


builder/azure/arm/builder.go
```diff
@@ -241,10 +241,6 @@ func (b *Builder) Run(ctx context.Context, ui packersdk.Ui, hook packersdk.Hook)
 		b.config.storageAccountBlobEndpoint = blobPrimaryEndpoint
 		// Remove trailing slash
 		azureClient.GiovanniBlobClient.Client.BaseUri = blobPrimaryEndpoint[:len(blobPrimaryEndpoint)-1]
-
-		if !equalLocation(account.Location, b.config.Location) {
-			return nil, fmt.Errorf("The storage account is located in %s, but the build will take place in %s. The locations must be identical", account.Location, b.config.Location)
-		}
```
tanmay-hc: Small doubt here. I might be wrong so please help me. This is what I found on the web.

https://learn.microsoft.com/en-us/azure/virtual-machines/faq-for-disks?tabs=azure-portal#can-i-use-a-vhd-file-in-an-azure-storage-account-to-create-a-managed-disk-in-a-different-region-

arsdragonfly: that doc is for VHD -> Managed Disk; we are doing VM Managed Disk SAS -> VHD.

arsdragonfly:
@tanmay-hc ready for another round of review

arsdragonfly:
@tanmay-hc sorry pushed to a different wrong branch earlier... now it's updated

PR #616, opened 2026-05-14
fix(winrm): respect winrm_use_ntlm setting instead of unconditionally forcing NTLM

### LINKED ISSUE #617: WinRM: winrm_use_ntlm setting is ignored -- NTLM transport unconditionally forced
## Description

The `setWinRMCertificate()` function in both the ARM and DTL builders unconditionally sets `WinRMTransportDecorator` to `&winrm.ClientNTLM{}`, completely ignoring the `winrm_use_ntlm` configuration option. This means **every** WinRM connection from the Azure plugin uses NTLM authentication regardless of user configuration.

### Affected Code

- `builder/azure/arm/config.go` line 968-972
- `builder/azure/dtl/config.go` line 492-496

```go
func setWinRMCertificate(c *Config) error {
    c.Comm.WinRMTransportDecorator =
        func() winrm.Transporter {
            return &winrm.ClientNTLM{}
        }
    // ...
}
```

The `packer-plugin-sdk` already handles this correctly in `communicator.Config.Prepare()` ([config.go line 638-640](https://github.com/hashicorp/packer-plugin-sdk/blob/main/communicator/config.go#L638-L640)), only setting the NTLM transport decorator when `WinRMUseNTLM` is true. However, the plugin's `setWinRMCertificate()` runs **before** `Comm.Prepare()` and unconditionally overrides this behavior.

## Impact

This causes **WinRM connection failures on hardened Windows images** where NTLM is rejected by the OS.

Specifically, on Windows Server 2022 with:
- **FIPS mode enabled** (`fips-mode-setup --enable` equivalent via `bcdedit /set {current} fipsenabled yes`)
- **DISA STIG hardening applied** (including HTTP.sys Extended Protection for Authentication)

HTTP.sys enforces Extended Protection for Authentication (EPA) with Channel Binding Tokens (CBT). The [`go-ntlmssp`](https://github.com/bodgit/ntlmssp) library used by `winrm.ClientNTLM` does not implement CBT support, so the NTLM Type 3 (AUTHENTICATE) message is rejected by HTTP.sys with HTTP 401.

### Error Symptoms

Packer builds against hardened Windows images fail with repeated WinRM connection errors:

```
azure-arm.windows: Retrying WinRM connection...
azure-arm.windows: WinRM connection error: unknown error Post "https://<ip>:5986/wsman": http response error: 401 -
```

Windows Security Event Log shows Event ID 4625 (logon failure) with:
- **Status**: `0xC000006D` (STATUS_LOGON_FAILURE)
- **Sub Status**: `0xC000006A` (STATUS_WRONG_PASSWORD) -- misleading; the password is correct but the NTLM token is rejected due to missing CBT

### Workaround

There is **no workaround** within the plugin configuration because `winrm_use_ntlm = false` (the default) is ignored. Users must either:
1. Skip Packer WinRM entirely and use `az vm run-command invoke` for post-build validation
2. Build a patched version of the plugin

## Expected Behavior

When `winrm_use_ntlm` is not set (default `false`), the plugin should use Basic auth over HTTPS, which works correctly because TLS provides transport encryption. The NTLM transport decorator should only be set when the user explicitly enables it with `winrm_use_ntlm = true`.

## Reproduction

1. Build a Windows Server 2022 golden image with FIPS mode and DISA STIG hardening
2. Create a Packer template using `azure-arm` builder with WinRM communicator:
   ```hcl
   communicator   = "winrm"
   winrm_use_ssl  = true
   winrm_insecure = true
   winrm_username = "packer"
   winrm_password = "..."
   # winrm_use_ntlm is NOT set (defaults to false)
   ```
3. Run `packer build` against the hardened image
4. Observe repeated HTTP 401 failures despite correct credentials

## Environment

- **Packer**: 1.11.2+
- **packer-plugin-azure**: v2.6.1
- **Target OS**: Windows Server 2022 with FIPS + DISA STIG
- **packer-plugin-sdk**: v0.6.7

## Testing Performed

We validated the fix with:

1. **Unit tests** (all pass):
   - `TestWinRMConfigShouldSetRoundTripDecoratorWhenNTLMEnabled` -- confirms NTLM decorator IS set when `winrm_use_ntlm = true`
   - `TestWinRMConfigShouldNotSetRoundTripDecoratorWhenNTLMDisabled` -- confirms NTLM decorator is NOT set when `winrm_use_ntlm` is unset
   - Tests pass for both ARM and DTL builders

2. **Integration testing** against production FIPS/STIG-hardened Windows Server 2022 images on Azure:
   - Confirmed NTLM Type 3 rejection (HTTP 401) with Event ID 4625 in Security log
   - Confirmed Basic auth over HTTPS (port 5986) succeeds when NTLM is not forced
   - Validated via `az vm run-command invoke` that the image itself is healthy (GA running, FIPS enabled, services nominal)

## Fix

PR #616 wraps the NTLM transport decorator in a conditional: `if c.Comm.WinRMUseNTLM { ... }`

sclark007:
## Summary

`setWinRMCertificate()` in both the ARM and DTL builders unconditionally sets `WinRMTransportDecorator` to `&winrm.ClientNTLM{}`, ignoring the `winrm_use_ntlm` configuration option. This makes it impossible to use Basic auth over HTTPS even when `winrm_use_ntlm` is not set (the default).

## Problem

On Windows Server 2022 images with FIPS mode enabled and DISA STIG hardening applied, HTTP.sys enforces Extended Protection for Authentication (EPA) with Channel Binding Tokens (CBT). The `go-ntlmssp` library used by `ClientNTLM` does not support CBT, causing the NTLM Type 3 authentication message to be rejected with HTTP 401.

Because the plugin unconditionally wraps WinRM with `ClientNTLM`, there is no way to fall back to Basic auth over HTTPS -- which works correctly since TLS provides the transport encryption.

The `packer-plugin-sdk` already handles this correctly in `communicator.Config.Prepare()` (line 638-640 of `config.go`), only setting the NTLM transport when `WinRMUseNTLM` is true. However, the plugin's `setWinRMCertificate()` runs before `Comm.Prepare()` and overrides this behavior.

## Fix

Wrap the NTLM transport decorator assignment in a conditional check:

```go
if c.Comm.WinRMUseNTLM {
    c.Comm.WinRMTransportDecorator = func() winrm.Transporter {
        return &winrm.ClientNTLM{}
    }
}
```

This aligns the plugin behavior with the SDK and allows users to opt out of NTLM when their target Windows images reject it.

## Testing

- Updated existing `TestWinRMConfigShouldSetRoundTripDecorator` to explicitly set `winrm_use_ntlm = true`
- Added new `TestWinRMConfigShouldNotSetRoundTripDecoratorWhenNTLMDisabled` verifying the decorator is nil when `winrm_use_ntlm` is not set
- All tests pass for both ARM and DTL builders

## Affected Builders

- `azure-arm` (`builder/azure/arm/config.go`)
- `azure-dtl` (`builder/azure/dtl/config.go`)

Fixes #617

### DIFF
--- a/builder/azure/arm/config.go
+++ b/builder/azure/arm/config.go
@@ -966,10 +966,12 @@ func setSshValues(c *Config) error {
 }
 
 func setWinRMCertificate(c *Config) error {
-	c.Comm.WinRMTransportDecorator =
-		func() winrm.Transporter {
-			return &winrm.ClientNTLM{}
-		}
+	if c.Comm.WinRMUseNTLM {
+		c.Comm.WinRMTransportDecorator =
+			func() winrm.Transporter {
+				return &winrm.ClientNTLM{}
+			}
+	}
 
 	cert, err := c.createCertificate()
 
--- a/builder/azure/arm/config_test.go
+++ b/builder/azure/arm/config_test.go
@@ -665,7 +665,25 @@ func TestConfigShouldSupportPackersConfigElements(t *testing.T) {
 	}
 }
 
-func TestWinRMConfigShouldSetRoundTripDecorator(t *testing.T) {
+func TestWinRMConfigShouldSetRoundTripDecoratorWhenNTLMEnabled(t *testing.T) {
+	config := getArmBuilderConfiguration()
+	config["communicator"] = "winrm"
+	config["winrm_username"] = "username"
+	config["winrm_password"] = "Password123"
+	config["winrm_use_ntlm"] = true
+
+	var c Config
+	_, err := c.Prepare(config, getPackerConfiguration())
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	if c.Comm.WinRMTransportDecorator == nil {
+		t.Error("Expected WinRMTransportDecorator to be set when winrm_use_ntlm is true, but it was nil")
+	}
+}
+
+func TestWinRMConfigShouldNotSetRoundTripDecoratorWhenNTLMDisabled(t *testing.T) {
 	config := getArmBuilderConfiguration()
 	config["communicator"] = "winrm"
 	config["winrm_username"] = "username"
@@ -677,8 +695,8 @@ func TestWinRMConfigShouldSetRoundTripDecorator(t *testing.T) {
 		t.Fatal(err)
 	}
 
-	if c.Comm.WinRMTransportDecorator == nil {
-		t.Error("Expected WinRMTransportDecorator to be set, but it was nil")
+	if c.Comm.WinRMTransportDecorator != nil {
+		t.Error("Expected WinRMTransportDecorator to be nil when winrm_use_ntlm is not set, but it was set")
 	}
 }
 
--- a/builder/azure/dtl/config.go
+++ b/builder/azure/dtl/config.go
@@ -490,10 +490,12 @@ func setSshValues(c *Config) error {
 }
 
 func setWinRMCertificate(c *Config) error {
-	c.Comm.WinRMTransportDecorator =
-		func() winrm.Transporter {
-			return &winrm.ClientNTLM{}
-		}
+	if c.Comm.WinRMUseNTLM {
+		c.Comm.WinRMTransportDecorator =
+			func() winrm.Transporter {
+				return &winrm.ClientNTLM{}
+			}
+	}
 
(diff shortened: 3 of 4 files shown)


tanmay-hc (approved):
LGTM. Thanks for the fix 🚀

PR #625, opened 2026-05-31
feat: add hostname-aware NSG rules for existing and outbound VNet paths

### LINKED ISSUE #468: `private_virtual_network_with_public_ip` doesn't allow security group to be applied to the public IP
#### Overview of the Issue

We are trying to build VMs in an existing VNet with private access to certain other resources required during build time. I expected, given that I enabled `private_virtual_network_with_public_ip` that the security group could be applied to limit ingress to the public IP, however instead on launch I am told:

```
  Error: 1 error(s) occurred:
 * If virtual_network_name is specified, allowed_inbound_ip_addresses cannot be specified
```

#### Reproduction Steps

`private_virtual_network_with_public_ip = true` and ensure you set `virtual_network_name` and `allowed_inbound_ip_addresses` to otherwise-valid values

### Plugin and Packer version

Packer 1.10.1
Azure plugin 2.2.1

```hcl
source "azure-arm" "azure-arm" {
  allowed_inbound_ip_addresses = ["${var.allowed_inbound_ip_address}"]
  async_resourcegroup_delete = false
  azure_tags = {
    "Created By"          = "${var.created_by}"
    "Git Branch"          = "${var.git_branch}"
    "Git Commit"          = "${var.git_commit}"
    Name                  = "Packer Builder - ${var.image_description}"
  }
  client_id                         = "guid-guid-guid"
  client_secret                     = "${var.azure_ad_application_client_secret}"
  communicator                      = "ssh"
  image_publisher                   = "RedHat"
  image_offer                       = "RHEL"
  image_sku                         = "810-gen2"
  location                          = "Australia East"
  
  private_virtual_network_with_public_ip = true
  virtual_network_name                   = "Packer-Builds"
  
  os_disk_size_gb                   = 200
  os_type                           = "Linux"

    gallery_name         = "Images"
    image_name           = "RHEL8"
    image_version        = "${var.build_version_short}"
    replication_regions  = ["Australia East"]
    resource_group       = "Images"
    subscription         = "guid-guid-guid"
    storage_account_type = "Standard_ZRS"
  }
  shared_gallery_image_version_exclude_from_latest = var.is_prerelease_build ? "true" : "false"

  subscription_id = "guid-guid-guid"
  tenant_id       = "guid-guid-guid"
  vm_size         = "Standard_D2a_V4"
  ssh_username    = "admin"
}
```

Invoked from Ubuntu 22 within Kubernetes, launching a RHEL base image (however it can't even launch the VM due to the validation failure). Public Azure cloud, Australia East region.

```
  Error: 1 error(s) occurred:
 * If virtual_network_name is specified, allowed_inbound_ip_addresses cannot be specified
```

### LINKED ISSUE #622: feat: support `allowed_inbound_ip_addresses` with existing virtual network
#### Community Note

Please vote on this issue by adding a thumbs-up reaction to the original issue if it is later opened upstream. Please do not add "+1" or "me too" comments.

#### Description

Support `allowed_inbound_ip_addresses` when `virtual_network_name` is used with a pre-existing VNet.

The Azure ARM builder supports `allowed_inbound_ip_addresses` only when it creates the temporary VNet and subnet itself. In that path, the builder creates a temporary Network Security Group (NSG), adds an inbound allow rule for the communicator port, and associates that NSG with the subnet used by the build VM.

When you set `virtual_network_name`, the builder switches to the pre-existing network path and rejects `allowed_inbound_ip_addresses` during configuration validation.

That limitation blocks a common enterprise case. Teams run image builds inside an existing VNet because they need private routing, internal DNS, private endpoints, egress controls, or access to internal package repositories. Those teams also need to restrict inbound access to the temporary build VM.

The limitation comes from the current implementation. The code associates the generated NSG with the subnet resource in the ARM template. That path works only when the builder owns the subnet resource in the template. Once you set `virtual_network_name`, the builder no longer owns that subnet resource, so the current subnet-association code cannot run.

This issue asks for the same allowlist behavior on an existing VNet without forcing the builder to modify shared subnet infrastructure.

Why this matters:

- It removes an inconsistency in builder behavior. The same setting should work whether the builder creates the network or uses an existing one.
- It lets teams keep a narrow inbound surface on temporary build VMs inside enterprise VNets.
- It avoids broader inbound exposure when users must set `virtual_network_name`.
- It fits locked-down environments where pre-existing networking is mandatory.
- It removes the need to pre-provision one-off NSG rules outside the build.

Any implementation of this feature must preserve the current behavior for builder-managed networks for backward compatibility. This feature request should not introduce a breaking change.

That approach keeps the current builder-managed path intact and avoids changes to shared subnets in existing VNets.

#### Use Case(s)

1. A build must run inside an existing corporate VNet so it can reach private package mirrors, internal artifact repositories, private endpoints, or internal DNS.

2. The build VM still needs a public IP or otherwise reachable communicator path, but only from a small set of source IPs such as a CI runner, jump host, or NAT gateway.

3. Security policy requires explicit inbound allowlists for temporary infrastructure, including image builds.

4. Teams want the builder to manage temporary build-time access control without permanent or manually maintained NSG rules on shared subnets.

5. Teams use `private_virtual_network_with_public_ip` and want the same `allowed_inbound_ip_addresses` behavior they already get on builder-managed networks.

#### Potential configuration

No new configuration may be necessary if the builder can extend the existing setting cleanly.

```hcl
source "azure-arm" "example" {
  virtual_network_name                = "shared-build-vnet"
  virtual_network_subnet_name         = "image-build-subnet"
  virtual_network_resource_group_name = "network-rg"

  private_virtual_network_with_public_ip = true

  allowed_inbound_ip_addresses = [
    "203.0.113.10/32",
    "198.51.100.0/24",
  ]
}
```

Expected behavior:

- the builder accepts this configuration
- the builder creates a temporary NSG for the build
- the NSG allows inbound TCP traffic only on the communicator port
- for existing VNet scenarios, the NSG is associated with the temporary NIC rather than modifying the existing subnet
- the temporary NSG is deleted during cleanup with the rest of the temporary build resources

### LINKED ISSUE #623: feat: support FQDN values in `allowed_inbound_ip_addresses`
#### Community Note

Please vote on this issue by adding a thumbs-up reaction to the original issue if it is later opened upstream. Please do not add "+1" or "me too" comments.

#### Description

Support FQDN values in `allowed_inbound_ip_addresses`.

The current setting accepts only literal IP addresses and CIDR blocks. That works when you know the exact source address range in advance, but it breaks down when the system that needs to reach the build VM sits behind a DNS name whose backing IPs can change.

Many teams do not control a fixed inbound source IP. They reach the build VM from CI runners, hosted build agents, corporate proxies, remote access gateways, or outbound NAT layers that operators publish through DNS names instead of stable IP ranges. In those environments, users often know the hostname they must allow, but they do not know the complete IP list ahead of time.

This issue asks for `allowed_inbound_ip_addresses` to accept FQDN values alongside literal IP addresses and CIDR blocks.

Any implementation of this feature must preserve the current behavior for existing IP and CIDR inputs for backward compatibility. This feature request should not introduce a breaking change.

Azure NSGs do not support FQDNs or CNAMEs in source rules. If the user supplies an FQDN, the builder must resolve it through normal DNS lookup, follow any CNAME chain, and add the resulting IP addresses to the allowlist.

That behavior matters because allowing only one returned address would create an incomplete and unreliable rule set. A hostname can map to several IPs for load balancing, regional routing, failover, or IPv4/IPv6 dual stack. The builder needs to allow the full set returned by DNS lookup at build time.

The feature also needs to handle CNAMEs correctly. If the supplied FQDN resolves through one or more CNAME records before DNS returns final A or AAAA answers, the builder must follow the normal DNS resolution chain and add the final returned IP addresses to the allowlist. The builder cannot place FQDNs or CNAMEs into the NSG because Azure does not support them.

Why this matters:

- It lets users express the real inbound source they know: a hostname, not a shifting IP list.
- It makes `allowed_inbound_ip_addresses` usable with hosted CI systems and enterprise gateways whose backing addresses change.
- It removes manual DNS-to-IP expansion outside the build.
- It reduces brittle configurations that break when operators rotate gateway IPs behind the same hostname.
- It keeps the security boundary narrow by expanding only the requested hostnames instead of forcing users to allow large CIDR ranges.

#### Use Case(s)

1. A team runs Packer from a hosted CI platform that publishes runner egress through a hostname or rotating gateway hostname rather than a single static IP.

2. A company routes developer or automation traffic through a secure outbound proxy or gateway whose address set changes over time, but whose DNS name stays stable.

3. A team uses `private_virtual_network_with_public_ip` and wants to allow inbound access from a corporate jump service that operators expose through DNS.

4. A platform team knows the hostname of an egress appliance or load-balanced remote access tier, but DNS returns several IPs and the full returned set must be allowed for reliable connectivity.

5. An environment publishes both IPv4 and IPv6 answers for the same hostname and the build must allow all resolved addresses that Azure NSG rules can support.

#### Potential configuration

No new setting may be necessary if the existing field can accept a mixed list of IPs, CIDRs, and FQDNs.

```hcl
source "azure-arm" "example" {
  allowed_inbound_ip_addresses = [
    "203.0.113.10/32",
    "198.51.100.0/24",
    "ci-gateway.example.com",
    "corp-proxy.example.net",
  ]
}
```

Expected behavior:

- the builder accepts literal IPs, CIDRs, and FQDNs in the same list
- for each FQDN, the builder performs DNS lookup before creating the NSG rule set
- the builder adds all returned IP addresses from DNS lookup to the NSG allowlist
- if DNS returns multiple addresses, the builder includes the full returned set rather than picking one address
- if DNS follows one or more CNAMEs before returning final A or AAAA answers, the builder includes the final returned IP addresses from that resolution chain
- the builder preserves the current communicator-port-only NSG behavior
- the builder preserves the current behavior for existing IP and CIDR inputs

### LINKED ISSUE #624: feat: support `deny_outbound_ip_addresses`
#### Community Note

Please vote on this issue by adding a thumbs-up reaction to the original issue if it is later opened upstream. Please do not add "+1" or "me too" comments.

#### Description

Support `deny_outbound_ip_addresses` for Azure ARM builds.

The current builder can narrow inbound access to the temporary build VM through `allowed_inbound_ip_addresses`, and the enhanced version under discussion would also support FQDN expansion for that allowlist. There is no corresponding way to deny selected outbound destinations from the temporary build VM during image creation.

That gap blocks a common enterprise image-build case. Some software performs first-boot registration, inventory enrollment, policy download, tenant binding, or environment-specific bootstrap when it can reach its backend during installation. During image creation, you often want the binaries in the image but you do not want the temporary build VM to register itself with a production or shared backend and bake that state into the captured image.

This is especially important for golden image workflows such as the Qualys Agent guidance for gold images, where the software must be present in the image but must not establish the wrong backend relationship during image creation. The same pattern appears with management agents, endpoint security tools, configuration agents, inventory collectors, monitoring agents, license managers, and bootstrap installers that contact service endpoints during install or first start.

This issue asks for a `deny_outbound_ip_addresses` setting that supports literal IP addresses, CIDR blocks, and FQDN values. It should act as the outbound counterpart to the enhanced `allowed_inbound_ip_addresses` behavior: instead of allowing selected inbound sources to the communicator port, the builder should create outbound deny rules that prevent the temporary build VM from initiating traffic to the specified destinations.

Azure NSGs do not support FQDNs or CNAMEs directly in destination rules. If the user supplies an FQDN, the builder must resolve it through normal DNS lookup, follow any CNAME chain, and add the resulting IP addresses to the outbound deny list.

Any implementation of this feature must preserve the current behavior for builds that do not use this new setting for backward compatibility. This feature request should not introduce a breaking change.

Why this matters:

- It lets teams install software during image creation without allowing accidental registration or enrollment against production backends.
- It keeps unwanted machine identity, tenant binding, configuration state, or licensing state out of the captured image.
- It makes golden image workflows safer for security agents, inventory agents, monitoring tools, and bootstrap installers.
- It reduces the need for external network controls or temporary DNS hacks just to keep one build VM away from a known backend.
- It gives users a builder-managed control that matches a real image-building need rather than forcing every team to solve the same problem outside Packer.
- It complements inbound restriction features by covering the opposite direction of network control during the build.

#### Use Case(s)

1. A team installs the Qualys Agent into a golden image and must prevent the temporary build VM from contacting the Qualys backend during image creation so that registration or configuration is not baked into the image.

2. A company installs endpoint security, monitoring, inventory, or configuration agents during the build, but those agents must not enroll with a production tenant until the final deployed VM boots in its real environment.

3. A platform team needs to block outbound calls from the build VM to a specific SaaS control plane, API endpoint, or licensing backend while still allowing the rest of the image build to proceed.

4. An application installer attempts automatic post-install bootstrap against a backend that would create environment-specific records, tokens, or configuration items that should belong only to real deployed machines, not the temporary image builder.

5. A team wants to deny communication to a short list of known destinations without having to route the whole build through separate firewall infrastructure.

6. A team knows the backend only by hostname, not by a stable IP list, so the deny list must support FQDNs and expand them to all returned addresses at build time.

7. A regulated environment needs explicit proof that the temporary build VM could not communicate with specific management or data exfiltration endpoints during image creation.

#### Potential configuration

If the builder can add this cleanly, one new field may be enough.

```hcl
source "azure-arm" "example" {
  communicator = "winrm"

  deny_outbound_ip_addresses = [
    "198.51.100.10/32",
    "203.0.113.0/24",
    "qagpublic.qg2.apps.qualys.com",
    "config-backend.example.net",
  ]
}
```

Expected behavior:

- the builder accepts literal IPs, CIDRs, and FQDNs in the same list
- for each FQDN, the builder performs DNS lookup before creating the NSG rule set
- the builder adds all returned IP addresses from DNS lookup to the outbound deny list
- if DNS follows one or more CNAMEs before returning final A or AAAA answers, the builder includes the final returned IP addresses from that resolution chain
- the builder creates NSG outbound deny rules that block the build VM from initiating traffic to the specified destinations
- the builder keeps normal inbound communicator behavior unchanged unless other settings modify it
- the builder preserves current behavior for builds that do not set `deny_outbound_ip_addresses`

tmeckel:
### Description

Adds three related Azure ARM networking improvements:
- allow allowed_inbound_ip_addresses when virtual_network_name is set by attaching the temporary NSG to the build NIC instead of the existing subnet
- allow allowed_inbound_ip_addresses and deny_outbound_ip_addresses to accept hostnames and expand them to literal IPs at build time
- add outbound NSG deny rules for deny_outbound_ip_addresses while keeping the no-field path unchanged

### Resolved Issues
If your PR resolves any open issue(s), please indicate them like this so they will be closed when your PR is merged:

Closes #622 
Closes #623 
Closes #624
Closes #468 

### Rollback Plan

If a change needs to be reverted, we will roll out an update to the code within 7 days.

Are there any changes to security controls (access controls, encryption, logging) in this pull request? If so, explain.


builder/azure/arm/template_factory.go
```diff
@@ -339,9 +354,14 @@ func GetVirtualMachineTemplateBuilder(config *Config) (*template.TemplateBuilder
...
-			err = builder.SetNetworkSecurityGroup(config.AllowedInboundIpAddresses, config.Comm.Port())
+			err = builder.SetNetworkSecurityGroup(expandedAllowedInboundIpAddresses, expandedDeniedOutboundIpAddresses, config.Comm.Port(), false)
+			if err != nil {
+				return nil, err
+			}
+		} else if (len(config.AllowedInboundIpAddresses) >= 1) || (len(config.DenyOutboundIpAddresses) >= 1) {
+			err = builder.SetNetworkSecurityGroup(expandedAllowedInboundIpAddresses, expandedDeniedOutboundIpAddresses, config.Comm.Port(), true)
 			if err != nil {
 				return nil, err
 			}
```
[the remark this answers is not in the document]
tmeckel: Fixed — the `PrivateVirtualNetworkWithPublicIp` branch in `template_factory.go:367` now attaches the NSG to the NIC even with empty allow/deny lists. Covered by the new `TestVirtualMachineDeployment_ExistingVNetPublicIP_AttachesNsgToNic` and the updated `TestVirtualMachineDeployment10` fixture.

builder/azure/arm/config.go
```diff
@@ -1670,7 +1674,61 @@ func assertAllowedInboundIpAddresses(ipAddresses []string, setting string) (bool
...
+					for _, r := range lastLabel {
+						if r < 'a' || r > 'z' {
+							lastIsAlpha = false
+							break
+						}
+					}
+					if lastIsAlpha {
+						return false, fmt.Errorf("The setting %s must only contain valid IP addresses, CIDR blocks, or hostnames", setting)
+					}
+				}
```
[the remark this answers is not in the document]
tmeckel: Fixed — dropped the `numericLabels == len(labels)-1` block. `1.2.3.example` and `1.example` are now accepted; `TestConfigShouldAllowHostnamesWithNumericLabels` covers it. `hasLetterInLastLabel` still rejects all-numeric TLDs.

builder/azure/common/template/template_builder.go
```diff
@@ -697,41 +721,132 @@ func (s *TemplateBuilder) deleteResourceDependency(resource *Resource, predicate
...
+func splitAddressesByFamily(addresses []string) map[string][]string {
+	families := map[string][]string{}
+	for _, address := range addresses {
+		family := "ipv6"
+		if strings.Contains(address, ".") {
+			family = "ipv4"
+		}
+		families[family] = append(families[family], address)
+	}
+	return families
```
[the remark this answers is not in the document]
tmeckel: Fixed — heuristic now keys off `':'`, so `::ffff:192.0.2.1` lands in IPv6. `TestSplitAddressesByFamily` covers it.

builder/azure/arm/address_expander.go
```diff
@@ -0,0 +1,131 @@
...
+
+package arm
+
+import (
+	"context"
+	"fmt"
+	"net"
+	"sort"
+	"strings"
+)
```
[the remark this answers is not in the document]
tmeckel: Addressed — `time` is imported and a 30s `defaultLookupTTL` is in place. Refactor also dropped `addressResolver`/`netResolver`/`LookupCNAME` for a `lookupIPAddrFunc` adapter, since `net.DefaultResolver.LookupIPAddr` resolves CNAMEs transitively.

builder/azure/arm/address_expander.go
```diff
@@ -0,0 +1,131 @@
...
+
+var defaultAddressResolver addressResolver = netResolver{}
+
+func (netResolver) LookupIPAddr(host string) ([]net.IPAddr, error) {
+	return net.DefaultResolver.LookupIPAddr(context.Background(), host)
+}
+
+func (netResolver) LookupCNAME(host string) (string, error) {
+	return net.DefaultResolver.LookupCNAME(context.Background(), host)
+}
```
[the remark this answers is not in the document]
tmeckel: Fixed — `expandMixedAddressList` derives `context.WithTimeout(context.Background(), 30s)` per lookup. `TestExpandMixedAddressList_PropagatesContextDeadline` asserts the deadline via `errors.Is`. The explicit CNAME traversal and its test are gone.

terraform/outputs.tf
```diff
@@ -0,0 +1,26 @@
...
+  value = local.resource_prefix
+}
+
+output "resource_suffix" {
+  value = local.resource_suffix
+}
+
+output "virtual_network_name" {
+  value = azurerm_virtual_network.vnet.name
+}
```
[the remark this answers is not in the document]
tmeckel: Fixed — added `output "virtual_network_subnet_name"` referencing `azurerm_subnet.subnet.name`.

docs-partials/builder/azure/arm/Config-not-required.mdx
```diff
@@ -384,9 +384,12 @@
 - `allowed_inbound_ip_addresses` ([]string) - Specify the list of IP addresses and CIDR blocks that should be
   allowed access to the VM. If provided, an Azure Network Security
   Group will be created with corresponding rules and be bound to
-  the subnet of the VM.
-  Providing `allowed_inbound_ip_addresses` in combination with
-  `virtual_network_name` is not allowed.
+  the subnet (builder VNet) or NIC (existing VNet).
```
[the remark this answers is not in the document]
tmeckel: Fixed — `allowed_inbound_ip_addresses` description now mentions hostnames and the build-time resolution behavior.

.web-docs/components/builder/arm/README.md
```diff
@@ -522,9 +522,12 @@ Providing `temp_resource_group_name` or `location` in combination with
 - `allowed_inbound_ip_addresses` ([]string) - Specify the list of IP addresses and CIDR blocks that should be
   allowed access to the VM. If provided, an Azure Network Security
   Group will be created with corresponding rules and be bound to
-  the subnet of the VM.
-  Providing `allowed_inbound_ip_addresses` in combination with
-  `virtual_network_name` is not allowed.
+  the subnet (builder VNet) or NIC (existing VNet).
```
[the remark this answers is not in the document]
tmeckel: Regenerated — ran `packer-sdc renderdocs` + `compile-to-webdocs.sh` to match the partial.

builder/azure/common/acceptance_helper.go
```diff
@@ -33,6 +41,9 @@ func CheckAcceptanceTestEnvVars(t *testing.T, params CheckAcceptanceTestEnvVarsP
 	if os.Getenv("ARM_SUBSCRIPTION_ID") == "" {
 		t.Fatalf("Test %s requires environment variable ARM_SUBSCRIPTION_ID is set", t.Name())
 	}
+	if os.Getenv("ARM_VIRTUAL_NETWORK_NAME") == "" {
+		t.Fatalf("Test %s requires environment variable ARM_VIRTUAL_NETWORK_NAME is set", t.Name())
+	}
```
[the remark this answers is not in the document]
tmeckel: Fixed — added `CheckVirtualNetworkName bool` to `CheckAcceptanceTestEnvVarsParams`. The 3 existing-VNet tests now opt in; the other 3 ACC tests no longer need `ARM_VIRTUAL_NETWORK_NAME`.

tmeckel:
@tanmay-hc Could you please run Copilot again on my changes after the first review? 🙏🏼

tmeckel:
@tanmay-hc Any chance to get the PR reviewed and merged?

builder/azure/arm/template_factory.go
```diff
@@ -339,9 +354,14 @@ func GetVirtualMachineTemplateBuilder(config *Config) (*template.TemplateBuilder
...
-	if (config.PublicIpSKU != "Basic") || (len(config.AllowedInboundIpAddresses) >= 1) {
+	if (config.PublicIpSKU != "Basic") || (len(config.AllowedInboundIpAddresses) >= 1) || (len(config.DenyOutboundIpAddresses) >= 1) {
 		if config.VirtualNetworkName == "" {
-			err = builder.SetNetworkSecurityGroup(config.AllowedInboundIpAddresses, config.Comm.Port())
+			err = builder.SetNetworkSecurityGroup(expandedAllowedInboundIpAddresses, expandedDeniedOutboundIpAddresses, config.Comm.Port(), false)
+			if err != nil {
+				return nil, err
+			}
+		} else if config.PrivateVirtualNetworkWithPublicIp || (len(config.AllowedInboundIpAddresses) >= 1) || (len(config.DenyOutboundIpAddresses) >= 1) {
+			err = builder.SetNetworkSecurityGroup(expandedAllowedInboundIpAddresses, expandedDeniedOutboundIpAddresses, config.Comm.Port(), true)
```
tanmay-hc, line 365: An existing VNet with `public_ip_sku = "Standard"` and no `allowed_inbound_ip_addresses` / `deny_outbound_ip_addresses` previously fell into the outer `if` (because `PublicIpSKU != "Basic"`), but the inner if `config.VirtualNetworkName == ""` was false, so it was silently skipped. That behaviour is preserved. However, the new condition also catches `PrivateVirtualNetworkWithPublicIp`, which was the workaround users used.
Before this PR that scenario had no NSG at all. Now it silently gets an allow-all inbound NSG on the NIC. This is a silent behaviour change for users who were relying on the existing VNet's own security controls.

Please document this explicitly and add a CHANGELOG entry. Consider gating it behind the explicit opt-in (`allowed_inbound_ip_addresses`) rather than inferring it from `private_virtual_network_with_public_ip`.

builder/azure/arm/address_expander.go
```diff
@@ -0,0 +1,91 @@
...
+	"sort"
+	"strings"
+	"time"
+)
+
+const defaultLookupTTL = 30 * time.Second
+
+type lookupIPAddrFunc func(ctx context.Context, host string) ([]net.IPAddr, error)
+
+var defaultAddressLookup lookupIPAddrFunc = net.DefaultResolver.LookupIPAddr
```
tanmay-hc, line 22: This is a mutable package-level variable. Tests monkey-patch it directly:
```
defaultAddressLookup = fakeLookup(...)
defer func() { defaultAddressLookup = net.DefaultResolver.LookupIPAddr }()
```
This is unsafe when tests run in parallel and is an anti-pattern for production code. Any concurrent call to `expandMixedAddressList` during a test will see the fake. Use a dependency-injection approach instead: pass the resolver into `GetVirtualMachineTemplateBuilder`, or expose it only via a `resolverFunc` argument to the relevant functions. The lookup parameter already exists on `expandMixedAddressList`, just thread it through from the call site in `template_factory.go`.

builder/azure/arm/template_factory.go
```diff
@@ -133,6 +133,21 @@ func GetVirtualMachineTemplateBuilder(config *Config) (*template.TemplateBuilder
...
+			return nil, err
+		}
+	}
+	expandedDeniedOutboundIpAddresses := config.DenyOutboundIpAddresses
+	if len(config.DenyOutboundIpAddresses) >= 1 {
+		expandedDeniedOutboundIpAddresses, err = expandMixedAddressList(config.DenyOutboundIpAddresses, nil)
+		if err != nil {
+			return nil, err
+		}
+	}
```
tanmay-hc, line 151: `expandMixedAddressList` is called inside `GetVirtualMachineTemplateBuilder`, which runs at build time. This means:
- Hostname validation at prepare time succeeds, but the build can fail at the time of template construction with a DNS error, giving the user a confusing "build failed" message rather than a clear validation error.
- The 30-second DNS timeout is hidden deep inside template construction, callers have no ability to set a context deadline or propagate cancellation.

Consider moving resolution into the `Run()` step (already has a context) or at least surfacing DNS errors through a dedicated pre-flight Prepare-phase check. At minimum, the function should accept a `context.Context from the caller rather than creating its own.

builder/azure/common/template/template_builder.go
```diff
@@ -697,41 +721,132 @@ func (s *TemplateBuilder) deleteResourceDependency(resource *Resource, predicate
...
-		(*resource.Properties.SecurityRules)[0].Properties.SourceAddressPrefixes = &srcIpAddresses
-	} else {
-		(*resource.Properties.SecurityRules)[0].Properties.SourceAddressPrefix = common.StringPtr("*")
+	if len(dstIpAddresses) > 0 {
+		outboundRules := makeOutboundSecurityRules(dstIpAddresses, 100)
+		if len(outboundRules) == 1 && len(*resource.Properties.SecurityRules) == 1 {
+			*resource.Properties.SecurityRules = append(*resource.Properties.SecurityRules, outboundRules...)
+		} else {
+			*resource.Properties.SecurityRules = append(outboundRules, *resource.Properties.SecurityRules...)
+		}
```
tanmay-hc, line 799: The intent is: when there is exactly one outbound rule and one inbound rule, append outbound after inbound. But the inbound rule's startingPriority is already bumped to 200 whenever `denyRuleCount > 0`. The "1-and-1" fast path results in the inbound rule having priority 200 and the outbound rule being appended at no explicit priority — it re-uses the same index slot, so in practice it is 200 as well, which is a duplicate priority that Azure will reject.

This logic should be replaced by an explicit list construction where priorities are always assigned sequentially and the outbound rules always come first (lower priority number = evaluated first in Azure NSG).

builder/azure/arm/config.go
```diff
@@ -1670,7 +1675,42 @@ func assertAllowedInboundIpAddresses(ipAddresses []string, setting string) (bool
...
+				hasLetterInLastLabel := false
+				for _, r := range lastLabel {
+					if r >= 'a' && r <= 'z' {
+						hasLetterInLastLabel = true
+						break
+					}
+				}
+				if !hasLetterInLastLabel {
+					return false, fmt.Errorf("The setting %s must only contain valid IP addresses, CIDR blocks, or hostnames", setting)
+				}
```
tanmay-hc, line 1746: `assertAllowedInboundIpAddresses` calls `normalizeHostname`, so validation and expansion already share the same normalisation. However, the validation rejects any TLD that contains no letter (e.g. a bare numeric TLD). The IANA rules allow all-numeric labels in non-TLD positions but the TLD itself must contain at least one letter — that part is correct. But it also rejects hostnames whose last label contains only uppercase letters (since it only checks r >= 'a' && r <= 'z'). Because normalizeHostname lower-cases the input first, running the character-set check on normalized is correct. But the function operates on ipAddress (the original, possibly mixed-case input) for the outer CIDR/IP check, then on normalized for the label check — this asymmetry is confusing. Add a comment or a single normalise-at-entry-point call to make this explicit.

builder/azure/common/acceptance_helper.go
```diff
@@ -53,3 +65,42 @@ func loggedIntoAzureCLI(t *testing.T) bool {
...
+		t.Fatal("api.ipify.org returned empty body. Cannot determine runner public IP.")
+	}
+
+	if net.ParseIP(ipStr) == nil {
+		t.Fatalf("api.ipify.org returned %q which is not a valid IP address.", ipStr)
+	}
+
+	if strings.Contains(ipStr, ":") {
+		return ipStr + "/128"
+	}
```
tanmay-hc, line 114: `DetectPackerPublicIP` makes an outbound HTTP call to `https://api.ipify.org` during acceptance tests. This adds a hard external dependency to the CI pipeline. If `api.ipify.org` is unreachable or rate-limits the runner, the test fails with a confusing message unrelated to the feature under test.

builder/azure/arm/config.go
```diff
@@ -1495,12 +1499,13 @@ func assertRequiredParametersSet(c *Config, errs *packersdk.MultiError) {
...
-				errs = packersdk.MultiErrorAppend(errs, err)
-			}
+		if ok, err := assertAllowedInboundIpAddresses(c.AllowedInboundIpAddresses, "allowed_inbound_ip_addresses"); !ok {
+			errs = packersdk.MultiErrorAppend(errs, err)
+		}
+	}
+	if len(c.DenyOutboundIpAddresses) >= 1 {
+		if ok, err := assertAllowedInboundIpAddresses(c.DenyOutboundIpAddresses, "deny_outbound_ip_addresses"); !ok {
+			errs = packersdk.MultiErrorAppend(errs, err)
 		}
```
tanmay-hc, line 1533: The function `assertAllowedInboundIpAddresses` is called to validate `DenyOutboundIpAddresses`. The validation logic itself is appropriate, but the name is wrong and will confuse future readers. Rename to something like `assertIPOrHostnameList` or `assertAddressList`, or extract a small shared validator.

tanmay-hc (changes requested):
## Minor Issues & Nits

### Redundant length guards around `expandMixedAddressList` calls

**File:** `builder/azure/arm/template_factory.go` lines 134–149

Both `expandMixedAddressList` calls are wrapped in `if len(...) >= 1` guards, but the function already handles an empty slice correctly and returns immediately. These guards add noise and imply the function cannot handle empty input. Either remove them or add a comment explaining the intent (e.g. to skip allocating a context).

---

### Two-level dedup in `expandMixedAddressList` needs a comment

**File:** `builder/azure/arm/address_expander.go` lines 51–63

Resolved IPs are deduplicated into a `localSeen` map per hostname, then again through the outer `seen` map in `appendUniqueAddress`. The behaviour is correct and tested, but without a comment a future reader may remove one of the two levels thinking it is redundant. A short note explaining that `localSeen` deduplicates within a single DNS response while `seen` deduplicates across all entries would make this clear.

---

### NSG rule name suffix capitalisation is inconsistent

**File:** `builder/azure/common/template/template_builder.go` line 847

`strings.ToUpper(family)` produces `IPV4`/`IPV6`, giving names like `AllowIPsToSshWinRMInboundIPV6`. The base name uses camelCase, so the all-caps suffix looks inconsistent. Consider `Ipv4`/`Ipv6` instead. More importantly, these names are visible in the deployed Azure NSG, so whatever the convention, add a comment marking them as a stable interface that must not change silently in a future refactor.

---

### `inboundRules` aliasing via pointer-to-slice

**File:** `builder/azure/common/template/template_builder.go` line 737

`inboundRules` is a plain slice stored as `SecurityRules: &inboundRules`. The subsequent `append(outboundRules, *resource.Properties.SecurityRules...)` may or may not reallocate, so `inboundRules` and `*resource.Properties.SecurityRules` could end up backed by different arrays. It works here because `inboundRules` is never used again, but it is fragile. Storing `SecurityRules` as a value type rather than a pointer-to-slice would eliminate this class of issue.

---

### Duplicate acceptance test template constants

**File:** `builder/azure/arm/builder_acc_test.go`

`testBuilderAccExistingVNetAllowedInboundIP` and `testBuilderAccAllowedInboundIpAddressesLiteral` appear to be identical — same VNet config, same allowlist placeholder, same image, same region; only `managed_image_name` differs. Either merge them into a single test or add a comment explaining what each is independently verifying.

---

### Missing test case for trailing-dot hostname

**File:** `builder/azure/arm/config.go`

`normalizeHostname` strips a trailing dot, so `host.` normalises to `host`, which is then correctly rejected by the `!strings.Contains(normalized, ".")` check. This is the right behaviour but it is implicit. Worth adding a small test case for `"host."` to make the intent explicit and prevent regression.

---

### No migration callout in docs for the lifted VNet restriction

**Files:** `.web-docs/components/builder/arm/README.md`, `website/docs/builders/azure/arm/Config-not-required.mdx`

The old docs stated _"Providing `allowed_inbound_ip_addresses` in combination with `virtual_network_name` is not allowed."_ The new docs simply remove that sentence. Users who structured their pipelines around this restriction deserve an explicit note, for example:

> Previously this combination was unsupported. As of this version it is supported — when an existing VNet is used the NSG is attached to the NIC rather than the subnet.

builder/azure/common/template/template_builder.go
tmeckel: Thanks for the review. After investigating against the Azure NSG API contract, I believe no code change is required here. Walking through your points:

**1. No duplicate priority exists.** Verified against the five approved templates that exercise the outbound-deny path (e.g. `TestVirtualMachineDeployment_WithOutboundDenyRule_UsesHigherPrecedenceThanBroadAllow.approved.json`):

- Inbound allow rule: `priority: 200` (bumped because `denyRuleCount > 0`)
- Outbound deny rule: `priority: 100` (explicit `startingPriority` passed to `makeOutboundSecurityRules`, then incremented per family via `priority++`)

These are distinct values, so Azure will not reject the deployment on duplicate-priority grounds.

**2. Array position ≠ priority slot.** The thread of reasoning conflates the slice index with the rule's `Priority` field. The priority is set at construction time (`newIngressSecurityRule`/`newOutboundSecurityRule`) and stored as a field on each rule; it is not derived from where the rule sits in the `securityRules` array. Azure does not infer priority from array order — see below.

**3. Azure evaluates by priority field, not array order.** Per [Network security groups overview](https://learn.microsoft.com/azure/virtual-network/network-security-groups-overview#security-rules) and the [NSG design guide](https://learn.microsoft.com/azure/networking/design-guide/network-application-security-groups#how-to-choose):

> Azure evaluates rules in priority order: lowest number (highest priority) first. When traffic matches a rule, processing stops.

The `securityRules` array on the NSG resource is treated as an unordered bag; the ARM template samples in [Relocate an NSG](https://learn.microsoft.com/azure/azure-resource-manager/management/relocation/relocation-virtual-network-nsg#prepare) and others confirm there is no sort contract on the array.

**4. Priority uniqueness is per-direction, not across the whole NSG.** Azure's own default rules reuse the same priority value across directions: `AllowVnetInBound` and `AllowVnetOutBound` are both priority `65000`; `DenyAllInbound` and `DenyAllOutbound` are both `65500`. So inbound priority 200 and outbound priority 100 are not in the same namespace and cannot collide.

**5. "Outbound rules always come first" is not required by Azure.** Lower priority number = evaluated first, yes — but "first" refers to evaluation order, not the position in the JSON array. Azure sorts internally. The current code already produces outbound-deny rules at priority 100 (lower than the bumped inbound allow at 200), so the deny is evaluated first within the outbound direction, which is the desired precedence.

Net: the "1-and-1" fast path does prepend or append based on counts, but since array order is not a behavioral input to Azure, the only effect is cosmetic ordering of the JSON. No duplicate priority, no rejection, no behavioral bug.

Happy to refactor for clarity if you still prefer an explicit construction order, but flagging this as a blocking issue is incorrect given the above.

tmeckel: Additional corroboration from the upstream Terraform provider: the `azurerm_network_security_group` resource in [terraform-provider-azurerm](https://github.com/hashicorp/terraform-provider-azurerm/blob/c05a5d12932cde25bfaf2742b227c527585b5d23/internal/services/network/network_security_group_resource.go) defines `security_rule` as a `pluginsdk.TypeSet` (see the schema near the top of `resourceNetworkSecurityGroup()`). `expandSecurityRules` (called in both Create and Update) iterates that set with `.List()` and appends each rule to a slice with no priority sort:

```go
func expandSecurityRules(d *pluginsdk.ResourceData) ([]networksecuritygroups.SecurityRule, error) {
    sgRules := d.Get("security_rule").(*pluginsdk.Set).List()
    rules := make([]networksecuritygroups.SecurityRule, 0)
    for _, sgRaw := range sgRules {
        ...
        rules = append(rules, networksecuritygroups.SecurityRule{
            Name:       pointer.To(sgRule["name"].(string)),
            Properties: &properties,
        })
    }
    return rules, nil
}
```

The resulting slice (in hash-set iteration order, which is deliberately non-deterministic) is assigned directly:

```go
payload.Properties.SecurityRules = pointer.To(sgRules)
```

and posted to the Azure NSG CreateOrUpdate endpoint. If Azure required the `securityRules` array to be sorted by priority, the canonical HashiCorp provider would be broken for every user — clearly it is not. This is consistent with point 3 above: Azure treats the array as an unordered bag and evaluates by the `priority` field.

builder/azure/arm/template_factory.go
tmeckel: Thanks for flagging this. You're right that this is a behavior change worth documenting explicitly.

**Why the change was made:** The `PrivateVirtualNetworkWithPublicIp` condition was added to resolve a real connectivity issue (raised by Copilot review in [discussion_r3340202432](https://github.com/hashicorp/packer-plugin-azure/pull/625#discussion_r3340202432), now resolved): when using an existing VNet with a **Standard SKU** public IP (the Azure default since Basic SKU is on the deprecation path) and no NSG, inbound traffic is blocked by default. This breaks the Packer communicator (SSH/WinRM) because no rule allows it. Before this PR, the plugin created no NSG in this scenario — it only worked if the user's existing subnet already had an NSG with an allow rule, or if they were using a Basic SKU public IP.

**What changed:** The plugin now creates an allow-all inbound NSG on the NIC when `private_virtual_network_with_public_ip = true` with an existing VNet, even when no explicit `allowed_inbound_ip_addresses` / `deny_outbound_ip_addresses` are set. This ensures the communicator can connect regardless of the existing subnet's NSG configuration.

**Why not gate behind explicit opt-in:** Gating only behind `allowed_inbound_ip_addresses` would reintroduce the Standard SKU connectivity break for users who don't set an allowlist. The Standard SKU is the default and Basic is deprecated, so this scenario affects most users with existing VNets.

**Documentation added:**
migration note added to the `private_virtual_network_with_public_ip` field description to `builder/azure/arm/config.go` and generated documentation, which updated `Config-not-required.mdx` and web docs README.

The temporary NSG is still removed during cleanup, so no persistent resources are left behind.

builder/azure/arm/template_factory.go
tmeckel: Good point — addressed with both a Prepare-phase pre-flight check and context threading.

**1. Prepare-phase pre-flight:** `Config.Prepare` now calls `validateHostnamesResolve` which attempts DNS resolution for each hostname in `allowed_inbound_ip_addresses` and `deny_outbound_ip_addresses`. Unresolvable hostnames surface as **warnings** (not hard errors, since DNS is inherently transient — a hostname may be temporarily unresolvable at prepare time but resolvable at build time). The warnings are returned in the `[]string` return value of `Prepare`, so Packer surfaces them to the user before the build starts.

**2. Context threading:** `expandMixedAddressList` now accepts a `context.Context` from the caller instead of creating its own `context.Background()`. The call chain is:

```
StepValidateTemplate.Run(ctx) / StepDeployTemplate.Run(ctx)
  → s.factory(ctx, config)         // templateFactoryFunc now takes ctx
  → GetVirtualMachineDeployment(ctx, config)
  → GetVirtualMachineTemplateBuilder(ctx, config)
  → expandMixedAddressList(ctx, entries, nil)
```

The `templateFactoryFunc` type signature changed from `func(*Config)` to `func(context.Context, *Config)`. The 30-second DNS timeout is now applied via `context.WithTimeout(ctx, defaultLookupTTL)` — it caps the caller's context rather than replacing it, so the caller's deadline propagates and cancellation works.

DNS errors at build time are still possible (hostnames may become unresolvable between Prepare and build), but they now surface with the caller's context and are caught early by the Prepare pre-flight warnings.

builder/azure/common/acceptance_helper.go
tmeckel: Addressed. `DetectPackerPublicIP` now checks the `PACKER_ACC_RUNNER_PUBLIC_IP` environment variable first. If set, it uses that literal IP directly — no outbound HTTP call to `api.ipify.org`. If the env var is not set, it falls back to the existing `api.ipify.org` call.

This lets CI inject the runner's public IP without the external dependency, while preserving the auto-detection path for local runs.

builder/azure/arm/address_expander.go
tmeckel: Thanks for calling this out. I agreed with the core point in your comment and reworked the implementation in that direction.

Current state of the implementation:
- tests no longer monkey-patch a mutable package-level resolver
- dependency injection is threaded from the template-factory call sites down into the address expansion helpers via an explicit `lookup lookupIPAddrFunc` parameter
- `expandMixedAddressList` already had a lookup parameter; that resolver is now passed through from `GetVirtualMachineTemplateBuilder` / deployment-construction call sites instead of being swapped globally in tests
- package-level `defaultAddressLookup` remains only as immutable production fallback when `lookup == nil`
- test coverage now uses per-test fake resolvers passed directly at the call site, so parallel tests do not share mutable resolver state

So this follows your suggestion: resolver is injected at relevant boundary rather than exposed as mutable shared state.

builder/azure/arm/config.go
tmeckel: Addressed in current implementation.

I made two changes to remove the asymmetry you pointed out:
- renamed the shared validator from `assertAllowedInboundIpAddresses` to `assertAddressList`, since it validates both `allowed_inbound_ip_addresses` and `deny_outbound_ip_addresses`
- normalized at entry to the validator (`normalized := normalizeHostname(address)`) and then used that normalized value consistently for IP/CIDR/hostname validation

So validation now follows one path for all address forms instead of mixing original input for some checks and normalized input for others. This keeps mixed-case hostnames explicit and easier to read.

builder/azure/arm/config.go
tmeckel: Fixed. The shared validator is now named `assertAddressList` and both call sites use that name:
- `allowed_inbound_ip_addresses`
- `deny_outbound_ip_addresses`

That matches actual behavior better and removes confusion from using an inbound-specific function name for outbound deny-list validation.

tmeckel:
@tanmay-hc Handled all your review comments. Code is ready for another review.

tanmay-hc:
@tmeckel Thank you for the PR and addressing all the comments in detail. The PR mostly looks good to me.
Two small things before I approve:
- In `builder_acc_test.go`, `testBuilderAccExistingVNetAllowedInboundIP` and `testBuilderAccAllowedInboundIpAddressesLiteral` appear to be identical. If they're testing the same scenario, could you consolidate them into one? If they're intentionally distinct, a short comment above each explaining the difference would help future readers avoid deleting one as a copy-paste mistake.
- Could you add a CHANGELOG entry for this PR? It covers a few distinct user-facing changes, and I feel you're best placed to capture those accurately.

tmeckel:
@tanmay-hc  Addressed. Those two ACC templates had drifted into the same existing-VNet + literal inbound allowlist scenario; the only material difference left was the managed image name.

I consolidated them into a single shared template and updated the existing-VNet literal allowlist test to use it, rather than keeping two copies that exercised the same path.

CHANGELOG has been updated

tanmay-hc (approved):
Thanks for the PR and incorporating the requested changes. The PR looks good to me.

PR #626, opened 2026-06-05
Adding support to create images with both source data disks and additional data disks

ojaswimittal2301:
### Description
Problem:
Packer lets users set disk_additional_size to create empty data disks. However, deployment fails when the source image already contains data disks, due to how the VM deployment ARM template is generated (LUN collisions / unset LUNs).

Change:
This PR allows building from a source that has data disks while also attaching additional empty data disks.

New optional parameter disk_additional_luns — lets users assign an explicit LUN to each disk in disk_additional_size.
Source data-disk LUN resolution — fetches existing data-disk LUNs from all three source types: Shared Image Gallery, Managed Image, and Platform/Marketplace images (PIR). The previous implementation resolved SIG and Managed Image sources but never issued a Get for PIR images, so a new VM Images client was added.
Template generation — combines user-specified additional LUNs and source LUNs to build a conflict-free VM deployment ARM template.
Fallback LUN assignment — when disk_additional_luns is omitted but disk_additional_size is set and the source has data disks, LUNs are auto-assigned by filling gaps between existing source LUNs. (lowest available lun)

Compatibility:
disk_additional_luns is optional; existing configurations are unaffected.

Validation:
Added config validation: disk_additional_luns must match the length of disk_additional_size, contain non-negative, non-duplicate values, and not conflict with source-image data-disk LUNs.

Testing:
Added unit tests.
Manually created a Shared Image Gallery version from a source containing a data disk and successfully attached additional data disks.

### Rollback Plan

If a change needs to be reverted, we will roll out an update to the code within 7 days.

Are there any changes to security controls (access controls, encryption, logging) in this pull request? If so, explain.


feiyushi (commented):
## Code Review

**TL;DR:** This PR fixes a real bug (additional disks used to silently overwrite source-image data disks) and adds a clean opt-in for explicit LUN assignment. Validation, tests, and docs are in good shape. **But it introduces a silent regression for the *implicit*-LUN code path** that any existing user with a data-disk-bearing source image + `disk_additional_size` will hit, and a couple of nil-deref / error-handling issues are worth tightening before merge.

---

### 🔴 Blocking issues

#### 1. Silent LUN collision when `disk_additional_luns` is **not** provided
`template_builder.SetAdditionalDisks` falls back to `Lun = IntPtr(i)` when the caller passes `luns == nil`:

```go
} else {
    additionalDataDisks[i].Lun = common.IntPtr(i)
}
```

But `existingSourceDataDisks` was just populated from the source image, which itself almost always starts at LUN 0. So a user who today has:

```hcl
image_publisher = "Canonical"; image_offer = "...with-data-disk"
disk_additional_size = [32]
```

…and never set `disk_additional_luns` (and never could — it doesn't exist on `main` yet) will, after this PR, get a template with two disks both at LUN 0. Azure rejects this and the user sees an opaque deployment error.

Pre-PR this worked (because `profile.DataDisks` was overwritten outright), so this is a **silent backward-compat regression** on the most common code path.

**Suggested fix:** when `luns == nil`, skip LUNs already used by `existingSourceDataDisks`:

```go
used := make(map[int]bool)
for _, d := range existingSourceDataDisks {
    if d.Lun != nil { used[*d.Lun] = true }
}
next := 0
// inside the loop, when luns is nil:
for used[next] { next++ }
additionalDataDisks[i].Lun = common.IntPtr(next)
used[next] = true
```

Please also add a `template_builder_test.go` case covering "source has LUN 0, additional disks supplied without explicit LUNs" — currently no test exercises the appended path.

#### 2. Possible nil dereference in `step_get_source_image_name.go` (SIG branch)

```go
if image.Properties != nil && image.Properties.StorageProfile.DataDiskImages != nil {
```

`StorageProfile` itself isn't nil-checked. The matching block you added in `resource_resolver.go` correctly checks all three (`Properties != nil && StorageProfile != nil && DataDisks != nil`) — please mirror that pattern here. (The pre-existing `Source.Id` check below has the same latent bug, but new code shouldn't extend it.)

---

### 🟡 Should-address

#### 3. PIR `Get` failure is silently swallowed
```go
pirImage, err := s.getPIRImage(ctx, skuVersionId)
if err != nil {
    s.say(fmt.Sprintf("Warning: Failed to fetch PIR image details for data disk LUN extraction: %s", err.Error()))
}
```
On error we *neither* halt *nor* mark `sourceImageDataDiskLuns` invalid. The downstream `SetAdditionalDisks` conflict check is now best-effort and the user can still produce a colliding template. Two reasonable options:

- Halt the step (cleanest — fail fast with a clear message).
- Surface a louder warning *and* skip the LUN-conflict guarantee message in docs.

Also: the PIR `Get` now fires on **every** PIR build, even ones without additional disks. Cheap, but worth gating on `len(s.config.AdditionalDiskSize) > 0` to avoid an extra round-trip.

#### 4. Cross-validation of user LUNs vs source LUNs only happens at template-build time
`assertRequiredParametersSet` validates duplicates and non-negativity, but the *interesting* collision (user LUN vs source LUN) is detected only deep inside `SetAdditionalDisks`. For PIR/SIG that's unavoidable, but for **`custom_managed_image_name`** the source LUNs are populated in `resourceResolver.Resolve` before validation completes — consider running the cross-check there so the user gets a config-time error rather than a build-time one.

#### 5. Inconsistent LUN type handling across the three branches
- PIR branch: `dd.Lun` is `*int`, nil-checked ✅
- SIG branch: `int32(dd.Lun)` (value type)
- Managed-image branch: `int32(dd.Lun)` (value type)

The three call sites all do essentially the same thing — please extract a helper (e.g., `extractLuns`) to deduplicate and make the type handling uniform.

---

### 🟢 Minor / nits

- `disk_additional_luns` field is documented as `([]int32)` — consider `(list of integers)` for user-facing docs consistency with other list fields.
- New `getPIRImage` injection follows the existing test-injection pattern ✅. Test stubs returning `(nil, nil)` are appropriate ✅.
- New `TestConfigAdditionalDiskLunsValidation` is thorough (negative, duplicate, length-mismatch, missing-sizes, plus 2 happy paths) ✅.
- No `*.approved.json` regenerations are needed in `template_builder_test.go` because the 3 existing call sites pass `nil` source disks and `nil` user LUNs, which makes `append(nil, additionalDataDisks...)` equivalent to the pre-PR `profile.DataDisks = &dataDisks` ✅. But please add at least one *new* approval test that exercises a non-empty source-disk list so this branch is snapshotted.
- `azure_client.go`: PIR client uses the same auth/UA wiring pattern as `galleryImagesClient` ✅.
- Consider documenting on the `AdditionalDiskLuns` field comment itself (not just the README) that source-image LUNs are reserved and any user LUN colliding with them will fail at build time.

---

### ✅ What's good
- Real bug fix (source-image disks were being silently dropped).
- Clean separation: new `SetSourceImageDataDisks` + reworked `SetAdditionalDisks` is easy to follow.
- Validation is exhaustive on the user-LUN list itself.
- Tests cover the new validation cases.
- Generated files (`config.hcl2spec.go`, `.web-docs/`, `docs-partials/`) regenerated correctly.
- License headers and code style are consistent with the rest of the codebase.

---

### Recommendation
**Request changes** — issues #1 and #2 should be fixed before merge; #3–#5 are strong nice-to-haves. The shape of the change is right and the bug fix is valuable, it just needs one more pass to avoid trading the old "silently drops source disks" bug for a new "silently builds a colliding template" bug on the implicit-LUN path.

ojaswimittal2301:
> ## Code Review
> **TL;DR:** This PR fixes a real bug (additional disks used to silently overwrite source-image data disks) and adds a clean opt-in for explicit LUN assignment. Validation, tests, and docs are in good shape. **But it introduces a silent regression for the _implicit_-LUN code path** that any existing user with a data-disk-bearing source image + `disk_additional_size` will hit, and a couple of nil-deref / error-handling issues are worth tightening before merge.
> 
> ### 🔴 Blocking issues
> #### 1. Silent LUN collision when `disk_additional_luns` is **not** provided
> `template_builder.SetAdditionalDisks` falls back to `Lun = IntPtr(i)` when the caller passes `luns == nil`:
> 
> ```go
> } else {
>     additionalDataDisks[i].Lun = common.IntPtr(i)
> }
> ```
> 
> But `existingSourceDataDisks` was just populated from the source image, which itself almost always starts at LUN 0. So a user who today has:
> 
> ```hcl
> image_publisher = "Canonical"; image_offer = "...with-data-disk"
> disk_additional_size = [32]
> ```
> 
> …and never set `disk_additional_luns` (and never could — it doesn't exist on `main` yet) will, after this PR, get a template with two disks both at LUN 0. Azure rejects this and the user sees an opaque deployment error.
> 
> Pre-PR this worked (because `profile.DataDisks` was overwritten outright), so this is a **silent backward-compat regression** on the most common code path.
> 
> **Suggested fix:** when `luns == nil`, skip LUNs already used by `existingSourceDataDisks`:
> 
> ```go
> used := make(map[int]bool)
> for _, d := range existingSourceDataDisks {
>     if d.Lun != nil { used[*d.Lun] = true }
> }
> next := 0
> // inside the loop, when luns is nil:
> for used[next] { next++ }
> additionalDataDisks[i].Lun = common.IntPtr(next)
> used[next] = true
> ```
> 
> Please also add a `template_builder_test.go` case covering "source has LUN 0, additional disks supplied without explicit LUNs" — currently no test exercises the appended path.
> 
> #### 2. Possible nil dereference in `step_get_source_image_name.go` (SIG branch)
> ```go
> if image.Properties != nil && image.Properties.StorageProfile.DataDiskImages != nil {
> ```
> 
> `StorageProfile` itself isn't nil-checked. The matching block you added in `resource_resolver.go` correctly checks all three (`Properties != nil && StorageProfile != nil && DataDisks != nil`) — please mirror that pattern here. (The pre-existing `Source.Id` check below has the same latent bug, but new code shouldn't extend it.)
> 
> ### 🟡 Should-address
> #### 3. PIR `Get` failure is silently swallowed
> ```go
> pirImage, err := s.getPIRImage(ctx, skuVersionId)
> if err != nil {
>     s.say(fmt.Sprintf("Warning: Failed to fetch PIR image details for data disk LUN extraction: %s", err.Error()))
> }
> ```
> 
> On error we _neither_ halt _nor_ mark `sourceImageDataDiskLuns` invalid. The downstream `SetAdditionalDisks` conflict check is now best-effort and the user can still produce a colliding template. Two reasonable options:
> 
> * Halt the step (cleanest — fail fast with a clear message).
> * Surface a louder warning _and_ skip the LUN-conflict guarantee message in docs.
> 
> Also: the PIR `Get` now fires on **every** PIR build, even ones without additional disks. Cheap, but worth gating on `len(s.config.AdditionalDiskSize) > 0` to avoid an extra round-trip.
> 
> #### 4. Cross-validation of user LUNs vs source LUNs only happens at template-build time
> `assertRequiredParametersSet` validates duplicates and non-negativity, but the _interesting_ collision (user LUN vs source LUN) is detected only deep inside `SetAdditionalDisks`. For PIR/SIG that's unavoidable, but for **`custom_managed_image_name`** the source LUNs are populated in `resourceResolver.Resolve` before validation completes — consider running the cross-check there so the user gets a config-time error rather than a build-time one.
> 
> #### 5. Inconsistent LUN type handling across the three branches
> * PIR branch: `dd.Lun` is `*int`, nil-checked ✅
> * SIG branch: `int32(dd.Lun)` (value type)
> * Managed-image branch: `int32(dd.Lun)` (value type)
> 
> The three call sites all do essentially the same thing — please extract a helper (e.g., `extractLuns`) to deduplicate and make the type handling uniform.
> 
> ### 🟢 Minor / nits
> * `disk_additional_luns` field is documented as `([]int32)` — consider `(list of integers)` for user-facing docs consistency with other list fields.
> * New `getPIRImage` injection follows the existing test-injection pattern ✅. Test stubs returning `(nil, nil)` are appropriate ✅.
> * New `TestConfigAdditionalDiskLunsValidation` is thorough (negative, duplicate, length-mismatch, missing-sizes, plus 2 happy paths) ✅.
> * No `*.approved.json` regenerations are needed in `template_builder_test.go` because the 3 existing call sites pass `nil` source disks and `nil` user LUNs, which makes `append(nil, additionalDataDisks...)` equivalent to the pre-PR `profile.DataDisks = &dataDisks` ✅. But please add at least one _new_ approval test that exercises a non-empty source-disk list so this branch is snapshotted.
> * `azure_client.go`: PIR client uses the same auth/UA wiring pattern as `galleryImagesClient` ✅.
> * Consider documenting on the `AdditionalDiskLuns` field comment itself (not just the README) that source-image LUNs are reserved and any user LUN colliding with them will fail at build time.
> 
> ### ✅ What's good
> * Real bug fix (source-image disks were being silently dropped).
> * Clean separation: new `SetSourceImageDataDisks` + reworked `SetAdditionalDisks` is easy to follow.
> * Validation is exhaustive on the user-LUN list itself.
> * Tests cover the new validation cases.
> * Generated files (`config.hcl2spec.go`, `.web-docs/`, `docs-partials/`) regenerated correctly.
> * License headers and code style are consistent with the rest of the codebase.
> 
> ### Recommendation
> **Request changes** — issues #1 and #2 should be fixed before merge; #3–#5 are strong nice-to-haves. The shape of the change is right and the bug fix is valuable, it just needs one more pass to avoid trading the old "silently drops source disks" bug for a new "silently builds a colliding template" bug on the implicit-LUN path.

1. added the suggested fallback. if additional data disk luns are not provided, we have the logic to set the luns instead of directly using iterator value. Logic is similar to the one decided for AIB. Start from 0 and keep skipping the values already used in source data disks.
2. Null check not added since image.Properties.StorageProfile is not a pointer
3. I dont think we should halt the step if the get PIR image call fails. I have instead modified the error statement to clearly state that the output may not contain source data disks 
4. It is better to maintain the same format for managed image, SIG and PIR
5. Did not exactly understand the comment. If I understand correctly, it suggests having a common function to get data disk luns from source image. But that won't be possible since the object structure is different for all cases (SIG, MDI, PIR)

feiyushi:
Thanks for the detailed responses, @ojaswimittal2301 — went through each point against the current code (`b3698d5`). Summary: **#1 and #2 are resolved**, #3–#5 I'm happy to concede. One small leftover on #1.

### 1. Implicit-LUN fallback — ✅ resolved
Confirmed the `nextLun`-skipping logic in `SetAdditionalDisks` now reserves source-image LUNs before auto-assigning, so the implicit path no longer collides on LUN 0. This matches the AIB behavior you referenced. 👍

One thing still open from the original comment: there's **no test exercising the appended path** (non-empty source LUNs + additional disks without explicit LUNs). The three existing `template_builder_test.go` calls all pass `nil, nil`, so the new merge/skip branch isn't snapshotted. Could you add one case like:

```go
_ = testSubject.SetSourceImageDataDisks([]int32{0})
err = testSubject.SetAdditionalDisks([]int32{32, 64}, nil, []int32{0}, "datadisk", compute.CachingTypesReadWrite)
// expect additional disks land on LUN 1 and 2, source disk stays on 0
```

This is the highest-risk logic in the PR and currently untested — worth locking in.

### 2. Nil check on `StorageProfile` — ✅ correct, conceded
You're right. Verified in the SDK: `GalleryImageVersionProperties.StorageProfile` is a value type (`StorageProfile GalleryImageVersionStorageProfile`, no `omitempty`/pointer), so it can't be nil and my suggestion doesn't apply here. No change needed. (The `resource_resolver.go` `DataDisks != nil` check is still correct there because that one *is* a pointer.)

### 3. PIR `Get` failure handling — 🟡 acceptable
Fair call not to halt; the clearer warning message is a good compromise given marketplace images with data disks are uncommon. One residual edge: if the PIR `Get` fails **and** the user supplied explicit `disk_additional_luns`, the source-vs-user LUN conflict guard silently can't run, so a colliding template could still deploy and fail opaquely at Azure. Not blocking — just worth a one-line note in the field docs that source-LUN reservation is best-effort for marketplace images. Also did you end up gating the `Get` on `len(disk_additional_size) > 0`? Minor, but avoids a round-trip on every PIR build.

### 4. Config-time cross-validation for managed images — conceded
Consistency across the three source types is a reasonable reason to keep it at build-time. Fine to leave as-is.

### 5. Shared `extractLuns` helper — conceded
Agreed — the three SDK shapes differ (`GalleryDataDiskImage.Lun int64`, managed `DataDisk.Lun int64`, PIR `DataDiskImage.Lun *int64`), so a single helper would need per-type adapters for marginal benefit. Not worth it.

---

Net: only the **#1 test** is a real ask before merge; everything else is resolved or optional. Nice work on the fix. 🚀

ojaswimittal2301:
> Thanks for the detailed responses, @ojaswimittal2301 — went through each point against the current code (`b3698d5`). Summary: **#1 and #2 are resolved**, #3–#5 I'm happy to concede. One small leftover on #1.
> 
> ### 1. Implicit-LUN fallback — ✅ resolved
> Confirmed the `nextLun`-skipping logic in `SetAdditionalDisks` now reserves source-image LUNs before auto-assigning, so the implicit path no longer collides on LUN 0. This matches the AIB behavior you referenced. 👍
> 
> One thing still open from the original comment: there's **no test exercising the appended path** (non-empty source LUNs + additional disks without explicit LUNs). The three existing `template_builder_test.go` calls all pass `nil, nil`, so the new merge/skip branch isn't snapshotted. Could you add one case like:
> 
> ```go
> _ = testSubject.SetSourceImageDataDisks([]int32{0})
> err = testSubject.SetAdditionalDisks([]int32{32, 64}, nil, []int32{0}, "datadisk", compute.CachingTypesReadWrite)
> // expect additional disks land on LUN 1 and 2, source disk stays on 0
> ```
> 
> This is the highest-risk logic in the PR and currently untested — worth locking in.
> 
> ### 2. Nil check on `StorageProfile` — ✅ correct, conceded
> You're right. Verified in the SDK: `GalleryImageVersionProperties.StorageProfile` is a value type (`StorageProfile GalleryImageVersionStorageProfile`, no `omitempty`/pointer), so it can't be nil and my suggestion doesn't apply here. No change needed. (The `resource_resolver.go` `DataDisks != nil` check is still correct there because that one _is_ a pointer.)
> 
> ### 3. PIR `Get` failure handling — 🟡 acceptable
> Fair call not to halt; the clearer warning message is a good compromise given marketplace images with data disks are uncommon. One residual edge: if the PIR `Get` fails **and** the user supplied explicit `disk_additional_luns`, the source-vs-user LUN conflict guard silently can't run, so a colliding template could still deploy and fail opaquely at Azure. Not blocking — just worth a one-line note in the field docs that source-LUN reservation is best-effort for marketplace images. Also did you end up gating the `Get` on `len(disk_additional_size) > 0`? Minor, but avoids a round-trip on every PIR build.
> 
> ### 4. Config-time cross-validation for managed images — conceded
> Consistency across the three source types is a reasonable reason to keep it at build-time. Fine to leave as-is.
> 
> ### 5. Shared `extractLuns` helper — conceded
> Agreed — the three SDK shapes differ (`GalleryDataDiskImage.Lun int64`, managed `DataDisk.Lun int64`, PIR `DataDiskImage.Lun *int64`), so a single helper would need per-type adapters for marginal benefit. Not worth it.
> 
> Net: only the **#1 test** is a real ask before merge; everything else is resolved or optional. Nice work on the fix. 🚀

Added unit tests

ojaswimittal2301:
Hi @tanmay-hc, we are from azure image builder team and this PR is regarding the lun number bug we discussed previously via email:
([EXTERNAL] Re:  High Priority Ask: Customizing images with data disks)

Can you please help move this forward

builder/azure/common/template/template_builder.go
```diff
@@ -372,27 +372,79 @@ func (s *TemplateBuilder) SetDiskEncryptionWithPaaSKey(securityType *hashiVMSDK.
...
-		dataDisks[i].CreateOption = "Empty"
-		dataDisks[i].Caching = cachingType
+		additionalDataDisks[i].DiskSizeGB = common.Int32Ptr(additionalSize)
+		if additionalDataDiskLuns != nil && i < len(additionalDataDiskLuns) {
+			// Validate that the explicit LUN doesn't conflict with source image data disk LUNs
+			if existingDataDiskLuns[int(additionalDataDiskLuns[i])] {
+				return fmt.Errorf("disk_additional_luns value %d conflicts with source image data disk LUN", additionalDataDiskLuns[i])
+			}
+			additionalDataDisks[i].Lun = common.IntPtr(int(additionalDataDiskLuns[i]))
+		} else {
```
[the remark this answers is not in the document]
ojaswimittal2301: With the current validations in place, the user can either skip disk_additional_luns altogether or specify a LUN for every data disk — partial specification is rejected at config-validation time ([len(AdditionalDiskLuns) != len(AdditionalDiskSize)](vscode-file://vscode-app/c:/Users/ojaswimittal/AppData/Local/Programs/Microsoft%20VS%20Code/6928394f91/resources/app/out/vs/code/electron-browser/workbench/workbench.html) is an error).

As a result, SetAdditionalDisks always follows either the all-explicit path or the all-implicit path, never a combination of the two. So both issue 1 (an explicit LUN being reused by later implicit assignment) and issue 2 (a silently tolerated length mismatch) are unreachable through the ARM builder.

builder/azure/arm/step_get_source_image_name.go
```diff
@@ -107,6 +117,28 @@ func (s *StepGetSourceImageName) Run(ctx context.Context, state multistep.StateB
 		s.config.ImageSku,
 		s.config.ImageVersion)
 
+	if len(s.config.sourceImageDataDiskLuns) > 0 {
```
tanmay-hc: This branch is dead code for platform images. On the PIR path `sourceImageDataDiskLuns` is always empty here (it's only set in the gallery branch, which returns earlier, and in the managed-image resolver). So `getPIRImage` is never called and platform image source data disks are never captured. I think the guard should be `len(s.config.AdditionalDiskSize) > 0` instead. We only need the source LUNs when we're going to override dataDisks.

builder/azure/arm/step_get_source_image_name_test.go
```diff
@@ -145,13 +146,19 @@ func TestStepGetSourceImageName(t *testing.T) {
 				GeneratedData: &genData,
 				say:           ui.Say,
 				error:         func(e error) {},
+				getPIRImage: func(ctx context.Context, id virtualmachineimages.SkuVersionId) (*virtualmachineimages.VirtualMachineImage, error) {
+					return nil, nil
```
tanmay-hc: This mock returns nil, nil, so the platform-image data-disk path is never actually exercised, which is how the dead-guard bug above slipped through. 
Can we add a case where getPIRImage returns an image with DataDiskImages and assert the LUNs get picked up?

builder/azure/common/template/template_builder.go
```diff
@@ -372,27 +372,79 @@ func (s *TemplateBuilder) SetDiskEncryptionWithPaaSKey(securityType *hashiVMSDK.
 	return nil
 }
 
-func (s *TemplateBuilder) SetAdditionalDisks(diskSizeGB []int32, dataDiskname string, cachingType hashiVMSDK.CachingTypes) error {
+func (s *TemplateBuilder) SetSourceImageDataDisks(luns []int32) error {
```
tanmay-hc, line 375: These source disks are emitted with only `lun` + `createOption: FromImage`, no `managedDisk/storageAccountType`.
Have you confirmed against a real capture (platform/managed image with data disks + additional disks) that Azure accepts this, including with a DiskEncryptionSet? Want to make sure we're not relying on defaults that don't hold everywhere.

builder/azure/arm/step_get_source_image_name.go
ojaswimittal2301: Yes you are right, the guard should be en(s.config.AdditionalDiskSize) > 0. Corrected, thanks!

builder/azure/arm/step_get_source_image_name_test.go
ojaswimittal2301: Added tests

builder/azure/arm/step_get_source_image_name.go
```diff
@@ -69,6 +72,13 @@ func (s *StepGetSourceImageName) Run(ctx context.Context, state multistep.StateB
 			return multistep.ActionContinue
 		}
 
+		// Extract source image data disk LUNs for template building
+		if image.Properties != nil && image.Properties.StorageProfile.DataDiskImages != nil {
```
tanmay-hc, line 76: One gap left: the Shared Gallery branch that captures data-disk LUNs here is still untested, since both `vmSourcedSigImageVersion` and `managedImageSourcedSigImageVersion` mocks leave `StorageProfile.DataDiskImages` unset. Could you add a gallery case where it's populated and assert `sourceImageDataDiskLuns` gets set, mirroring the new PIR test? That would lock in all three capture paths (managed / SIG / PIR).

tanmay-hc: Hey @ojaswimittal2301 
Could you please address this? We plan to do a release soon and would like to take this feature in the same.

builder/azure/common/template/template_builder.go
ojaswimittal2301: I am now passing the managed disk values for source disk as well(so that des is set for source data disks as well). Deployed this and verified from our logs that the deployed template has the correct des set for all data disks. Request body sent for vm template deployment:
"dataDisks": [
        {
          "lun": 8,
          "createOption": "FromImage",
          "managedDisk": {
            "diskEncryptionSet": {
              "id": "/subscriptions/de212587-b40e-43d8-9185-1ef5b2519da4/resourceGroups/ojaswimittal_p2/providers/Microsoft.Compute/diskEncryptionSets/testdes"
            }
          }
        },
        {
          "lun": 0,
          "name": "pkrdd4a5kmo1t8j-1",
          "createOption": "Empty",
          "caching": "None",
          "managedDisk": {
            "diskEncryptionSet": {
              "id": "/subscriptions/de212587-b40e-43d8-9185-1ef5b2519da4/resourceGroups/ojaswimittal_p2/providers/Microsoft.Compute/diskEncryptionSets/testdes"
            }
          },
          "diskSizeGB": 10
        }
      ]

builder/azure/arm/step_get_source_image_name.go
ojaswimittal2301: Hi @tanmay-hc , sorry for the delay, there was a high-priority task that came up and had to be prioritised. I have added the test case now

tanmay-hc: @ojaswimittal2301 The tests are failing. Please check.

ojaswimittal2301: I have pushed the fix, waiting for tests to run

tanmay-hc:
@ojaswimittal2301 Please sign the CLA and we can merge the PR.

ojaswimittal2301:
> @ojaswimittal2301 Please sign the CLA and we can merge the PR.

Hi, I just signed the CLA, not sure why the validation is not passing, do we need to manually rerun the validation?

ojaswimittal2301:
Hi @tanmay-hc, I realised that I had used the incorrect email to push the commits, I am trying to rewrite the commits and force push? Is there some other way around this? I cannot link the email to this github account as that is already linked to my enterprise account. Do you recommend creating a fresh PR and copying these commits?

ojaswimittal2301:
@tanmay-hc, I have reauthored/rebased to fix the CLA. Please approve the workflow and review whenever you have a moment

tanmay-hc (approved):
Thank you for all the changes! PR looks good to me.

PR #636, opened 2026-07-02
Add LUN numbers as tags to data disk snapshots

ojaswimittal2301:
### Description

When the ARM builder snapshots a VM's data disks, the resulting snapshots did not record which LUN each disk was attached to. This makes it hard for downstream consumers to reconstruct the original disk-to-LUN layout when rehydrating disks from the snapshots.

This PR fetches the LUN alongside the managed disk resource ID for each data disk on the VM, then propagates the LUN onto the corresponding snapshot as a `lun` tag.

**What changed:**

- **`step_get_additional_disks.go`** — Introduced a `DataDiskInfo` struct (`Lun`, `ManagedDiskID`) and changed the step to read both the `Lun` and the managed disk ID from the VM's `StorageProfile.DataDisks`, storing `[]DataDiskInfo` in the state bag instead of `[]string`.
- **`step_snapshot_data_disks.go`** — For each data disk, the shared `ArmTags` map is cloned into a per-disk copy and a `lun` tag (the LUN as a string) is added before creating the snapshot. Cloning avoids mutating the shared tags map used by other steps.
- **`step_deploy_template.go`** — Updated the `Cleanup` consumer to work with the new `[]DataDiskInfo` type (uses `disk.ManagedDiskID`).
- **Tests** — Updated existing tests for the new type and added coverage:
  - `TestStepGetAdditionalDiskShouldCaptureLunAndManagedDiskID` verifies LUN + managed disk ID are captured for multiple disks with non-zero LUNs.
  - `TestStepSnapshotDataDisksShouldAddLunTagAndPreserveOriginalTags` verifies the per-disk `lun` tag is set correctly, original user tags are preserved, and the shared tags map is not mutated.
  
**Testing:**

Tested the changes locally by creating a managed image from a sig source. The snapshot contains lun number tag:
image


ojaswimittal2301:
Hi @tanmay-hc, we have another PR for data disk related work in azure, can you help review this as well

ojaswimittal2301:
Hi @tanmay-hc, gentle ping for this

builder/azure/arm/step_get_additional_disks.go
```diff
@@ -65,14 +72,17 @@ func (s *StepGetDataDisk) Run(ctx context.Context, state multistep.StateBag) mul
...
 	if vm.Properties.StorageProfile.DataDisks != nil {
-		var vhdUri string
-		additionalDisks := make([]string, len(*vm.Properties.StorageProfile.DataDisks))
-		for i, additionalDisk := range *vm.Properties.StorageProfile.DataDisks {
-			vhdUri = *additionalDisk.ManagedDisk.Id
-			s.say(fmt.Sprintf(" -> Managed Additional Disk %d  : '%s'", i+1, vhdUri))
-			additionalDisks[i] = vhdUri
+		dataDisks := make([]DataDiskInfo, len(*vm.Properties.StorageProfile.DataDisks))
+		for i, disk := range *vm.Properties.StorageProfile.DataDisks {
+			managedDiskID := *disk.ManagedDisk.Id
```
tanmay-hc, line 74: `disk.ManagedDisk` is a pointer and, per the Azure SDK struct definition, is tagged `omitempty`. If Azure ever returns a data disk record without a managed disk reference, this line will panic at runtime. The current code has the same issue, but since we're touching this loop, let's add a guard here.

builder/azure/arm/step_get_additional_disks.go
```diff
@@ -22,6 +22,13 @@ type StepGetDataDisk struct {
 	error  func(e error)
 }
 
+// DataDiskInfo holds information about a data disk attached to the VM,
+// including its LUN and managed disk resource ID.
+type DataDiskInfo struct {
```
tanmay-hc: nit: can we move this to `artifact.go` since its used at many other steps too?

builder/azure/arm/step_snapshot_data_disks.go
```diff
@@ -74,17 +74,24 @@ func (s *StepSnapshotDataDisks) Run(ctx context.Context, stateBag multistep.Stat
...
 	for i, disk := range additionalDisks {
-		s.say(fmt.Sprintf(" -> Data Disk   : '%s'", disk))
+		s.say(fmt.Sprintf(" -> Data Disk (LUN %d) : '%s'", disk.Lun, disk.ManagedDiskID))
+
+		// Clone tags and add LUN number
+		snapshotTags := make(map[string]string, len(tags)+1)
+		for k, v := range tags {
+			snapshotTags[k] = v
+		}
+		snapshotTags["lun"] = strconv.Itoa(int(disk.Lun))
```
tanmay-hc: If a user has already defined a tag named `lun` in their `azure_tags` configuration, this code silently overwrites it with the LUN number. This is a behavioural regression for anyone using that tag name today. The fix should be one of:

- Warn and override — log a message saying "lun" tag is reserved and will be set to the disk LUN.
- Use a namespaced key — e.g., `"packer:lun"` or `"packer_lun"` to avoid collisions with user-defined tags.
- Skip if already set — honour the user's value if present.

I feel option 2 might be the cleanest.

builder/azure/arm/step_snapshot_data_disks.go
```diff
@@ -74,17 +74,24 @@ func (s *StepSnapshotDataDisks) Run(ctx context.Context, stateBag multistep.Stat
...
 
 		dstSnapshotName := dstSnapshotPrefix + strconv.Itoa(i)
```
tanmay-hc, line 98: The snapshot name suffix is the loop index (0, 1, 2, …), while the new "lun" tag carries the actual LUN. If the motivation of this PR is to correlate snapshots to LUNs, using the LUN in the snapshot name (e.g., dstSnapshotPrefix + strconv.Itoa(int(disk.Lun))) would make that correlation self-evident without having to inspect tags. This is a pre-existing naming scheme and changing it may be out of scope, but it is worth noting for consistency with the stated goal.

builder/azure/arm/step_snapshot_data_disks.go
```diff
@@ -74,17 +74,24 @@ func (s *StepSnapshotDataDisks) Run(ctx context.Context, stateBag multistep.Stat
 	var resourceGroupName = stateBag.Get(constants.ArmManagedImageResourceGroupName).(string)
 	var location = stateBag.Get(constants.ArmLocation).(string)
 	var tags = stateBag.Get(constants.ArmTags).(map[string]string)
-	var additionalDisks = stateBag.Get(constants.ArmAdditionalDiskVhds).([]string)
+	var additionalDisks = stateBag.Get(constants.ArmAdditionalDiskVhds).([]DataDiskInfo)
```
tanmay-hc: Unlike `step_deploy_template.go` (line 213–215) which guards the same state bag key with a nil-check before asserting, this line does a bare type assertion directly on `stateBag.Get`. If `ArmAdditionalDiskVhds` was never put into the state bag, Get returns nil and the assertion panics. Use the same GetOk / nil-guard pattern used in `StepDeployTemplate.Cleanup` for consistency and safety.

tanmay-hc (changes requested):

builder/azure/arm/step_snapshot_data_disks.go
ojaswimittal2301: Yes makes sense, we did consider it during the design phase. But we thought it might be a breaking change for downstream dependencies this includes our service, our current implementation won't support this change

builder/azure/arm/step_snapshot_data_disks.go
ojaswimittal2301: moved to artifact.go

builder/azure/arm/step_get_additional_disks.go
ojaswimittal2301: added a null check

builder/azure/arm/step_snapshot_data_disks.go
ojaswimittal2301: added null check

ojaswimittal2301:
Hi @tanmay-hc, please take a look whenever possible

tanmay-hc (approved):

